PR #7, opened 2015-05-01
add parallel install of packages

analogrelay:
/cc @yishaigalatzer @pranavkm @emgarten

### DIFF
--- a/src/NuGet.CommandLine/CommandOutputLogger.cs
+++ b/src/NuGet.CommandLine/CommandOutputLogger.cs
@@ -10,6 +10,7 @@ namespace NuGet.CommandLine
     public class CommandOutputLogger : ILogger
     {
         private readonly CommandOutputLoggerProvider _provider;
+        private static readonly object _consoleLock = new object();
 
         public CommandOutputLogger(CommandOutputLoggerProvider commandOutputProvider)
         {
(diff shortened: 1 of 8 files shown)


src/NuGet.CommandLine/CommandOutputLogger.cs
```diff
@@ -10,6 +10,7 @@ namespace NuGet.CommandLine
     public class CommandOutputLogger : ILogger
     {
         private readonly CommandOutputLoggerProvider _provider;
+        private static object _consoleLock = new object();
```
pranavkm: `readonly`

src/NuGet.CommandLine/Program.cs
```diff
@@ -86,10 +88,31 @@ public int Main(string[] args)
                         project,
                         sources.Values.Select(s => new PackageSource(s)),
                         packagesDir);
+                    if (parallel.HasValue())
```
pranavkm, line 91: Could this work like a switch where it picks a default value (4 \ 8) if it doesn't have one? Would be nice to go
`restore --parallel` and have it use sensible defaults.

pranavkm, line 91: Oh nevermind. This is happening later on in the code.

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -42,14 +43,14 @@ public RestoreCommand(ILoggerFactory loggerFactory)
                 return new RestoreResult(success: false, restoreGraphs: Enumerable.Empty<RestoreTargetGraph>());
             }
 
-            var projectLockFilePath = Path.Combine(request.Project.BaseDirectory, LockFileFormat.LockFileName);
+            var projectLockFilePath = string.IsNullOrEmpty(request.LockFilePath) ?
```
pranavkm, line 46: A lot of the real benefits in `dnu restore` was primarily from restoring multiple project.json in parallel. We should figure that out as a secondary work item.

pranavkm:
:shipit:

PR #12, opened 2015-05-05
Replacing Microsoft.Framework.Logging with NuGet.Logging.

pranavkm:


pranavkm:
cc @emgarten

src/NuGet.Client/HttpSource.cs
```diff
@@ -1,289 +0,0 @@
-// Copyright (c) Microsoft Open Technologies, Inc. All rights reserved.
```
pranavkm, line 1: These are unused in the application and use the `Microsoft.Framework.ILogger`.

PR #13, opened 2015-05-05
* Removing unused references

pranavkm:
- Turn on warningAsErrors for src projects
- Fix warnings
- Move HashCodeCombiner to common.


emgarten:
Much better! :shipit:

PR #14, opened 2015-05-05
Updating ExternalProjectReference

emgarten:
This change adds a constructor to ExternalProjectReference so that it can be used as is from NuGet.Commands and passed into the RestoreRequest context.

//cc @pranavkm

### DIFF
--- a/src/NuGet.ProjectModel/ExternalProjectReference.cs
+++ b/src/NuGet.ProjectModel/ExternalProjectReference.cs
@@ -1,4 +1,6 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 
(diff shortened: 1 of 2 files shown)


src/NuGet.ProjectModel/ExternalProjectReference.cs
```diff
@@ -7,19 +9,69 @@ namespace NuGet.ProjectModel
...
+            _name = name;
+            _packageSpecPath = packageSpecPath;
+            _projectReferences = projectReferences.ToList().AsReadOnly();
+        }
+
         /// <summary>
         /// The name of the external project
         /// </summary>
-        public string Name { get; }
+        public string Name
```
pranavkm, line 47: Why do you hate C# 6?

pranavkm:
:shipit:

PR #19, opened 2015-05-07
Removing repo-initialize for faster restores

pranavkm:


emgarten:
When is initialize needed?

pranavkm:
It's the first step in the build process. We can defer to `k-standard-goals` to do the restore (which does it in parallel).

emgarten:
:shipit:

PR #20, opened 2015-05-08
Keep in sync with latest changes in dnu

davidfowl:
- Added resource patterns
- Fixed some bugs with TFM parser


src/NuGet.Client/ManagedCodeConventions.cs
```diff
@@ -215,16 +206,29 @@ internal ManagedCodePatterns(ManagedCodeConventions conventions)
                         "runtimes/{rid}/native/{any}",
                         "native/{any}",
                     });
+
+                ResourceAssemblies = ResourceAssemblies = new PatternSet(
```
analogrelay, line 210: How's your recovery from dental surgery going there @davidfowl ? :P

PR #21, opened 2015-05-08
Circumvent slow feeds if possible and made more things run in parallel

davidfowl:
- Use Task.WhenAny rather than WhenAll in case faster feeds have the exact match are preferred
- Made more things run in parallel


yishaigalatzer:
Can you add a test to verify that none exact matches don't win?

PR #22, opened 2015-05-08
Restore perf

davidfowl:


davidfowl:
VS keeps adding those service includes soo noise

pranavkm:
:shipit:

test/NuGet.DependencyResolver.Core.Tests/ResolverFacts.cs
```diff
@@ -0,0 +1,132 @@
...
+
+namespace NuGet.DependencyResolver.Core.Tests
+{
+    public class ResolverFacts
+    {
+        [Fact]
+        public void FasterProviderReturnsResultsBeforeSlowOnesIfExactMatchFound()
+        {
+            // A 
+            var slowProvider = new TestProvider(TimeSpan.FromSeconds(2));
```
analogrelay, line 20: Whaaaaat? Do we really need to keep running this 2sec test every time we build just to verify that we take the fastest result? This seems like a lot of busy work during every build.

PR #27, opened 2015-05-13
* Removing implicit cast to prevent cast exceptions.

pranavkm:
- RemoteDependencyWalker null refs when match is non-null but the library
- is
- Print the package path correctly.

[Review on Reviewable](https://reviewable.io/reviews/nuget/nuget3/27)

### DIFF
--- a/src/NuGet.CommandLine/Program.cs
+++ b/src/NuGet.CommandLine/Program.cs
@@ -21,7 +21,7 @@ namespace NuGet.CommandLine
         public int Main(string[] args)
         {
 #if DEBUG
-            if(args.Contains("--debug"))
+            if (args.Contains("--debug"))
             {
                 args = args.Skip(1).ToArray();
                 System.Diagnostics.Debugger.Launch();
@@ -75,7 +75,7 @@ namespace NuGet.CommandLine
                             var packagesDir = packagesDirectory.HasValue() ?
                                 packagesDirectory.Value() :
                                 Path.Combine(Environment.GetEnvironmentVariable("USERPROFILE"), ".dnx", "packages");
-                            _log.LogVerbose($"Using packages directory: {packagesDirectory}");
+                            _log.LogVerbose($"Using packages directory: {packagesDir}");
 
                             // Run the restore
                             var request = new RestoreRequest(
--- a/src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
+++ b/src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
@@ -311,10 +311,19 @@ namespace NuGet.DependencyResolver
             var tasks = new List<Task<RemoteMatch>>();
             foreach (var provider in providers)
             {
-                Func<Task<RemoteMatch>> taskWrapper = async () => new RemoteMatch
+                Func<Task<RemoteMatch>> taskWrapper = async () =>
                 {
-                    Provider = provider,
-                    Library = await action(provider)
+                    var library = await action(provider);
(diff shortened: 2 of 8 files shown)


src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -331,7 +331,7 @@ public Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library, NuGe
                 // If we found an exact match then use it.
                 // This allows us to shortcircuit slow feeds even if there's an exact match
                 if (!libraryRange.VersionRange.IsFloating &&
-                    match != null &&
+                    match?.Library != null &&
```
pranavkm: Ran into this when trying it out locally. We have a match but it doesn't have a library. Is that a bug in the invoking code? cc @davidfowl

pranavkm:
cc @anurse

src/NuGet.CommandLine/Program.cs
```diff
@@ -75,7 +75,7 @@ public int Main(string[] args)
                             var packagesDir = packagesDirectory.HasValue() ?
                                 packagesDirectory.Value() :
                                 Path.Combine(Environment.GetEnvironmentVariable("USERPROFILE"), ".dnx", "packages");
-                            _log.LogVerbose($"Using packages directory: {packagesDirectory}");
+                            _log.LogVerbose($"Using packages directory: {packagesDir}");
```
yishaigalatzer, line 79: format the file

src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
yishaigalatzer: Is this related to an unlisted package?

src/NuGet.ProjectModel/LockFileLibrary.cs
```diff
@@ -50,13 +50,14 @@ public class LockFileTargetLibrary
     }
     public class LockFileItem
     {
-        public string Path { get; set; }
```
yishaigalatzer, line 53: one class per file

yishaigalatzer:
:shipit: when comment are addressed and when you figured out the null Library thingie

src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
pranavkm: I didn't debug this - I ran into null refs here when trying to get the msbuild task running. @davidfowl did make this change recently so I wanted to check if he already knew of this.

davidfowl: Hmmm gotta check the dnu code, what provider matched?

pranavkm:
<sup>**[src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs, line 334 [r1]](https://reviewable.io:443/reviews/nuget/nuget3/27#-JpHpXiGvrwMEnqPDTCn-r1-334)** ([raw file](https://github.com/nuget/nuget3/blob/b9434fba0f6000317305e21801d77c43a8b77ecc/src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs#L334)):</sup>
Nevermind - found the issue. I was returning a match when the library was null a few lines above this. This is fixed.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/nuget/nuget3/27)

PR #29, opened 2015-05-14
Add support for build/ folder in transitive restore

analogrelay:
During transitive restore, we need to add support for MSBuild targets and props files in packages. We do this by generating a `[ProjectName].nuget.targets` and `[ProjectName].nuget.props` file on restore. This contains machine-specific imports from NuGet packages. It should not be checked in, but is rather a product of the restore process (and thus a build artifact).

Now, a project only needs to add a single pair of conditional `<Import>` elements to pull in these two files and transitive restore will ensure the rest of the data is pulled in as needed.

/cc @lodejard @davidfowl @emgarten @yishaigalatzer @pranavkm

[Review on Reviewable](https://reviewable.io/reviews/nuget/nuget3/29)


analogrelay:
<sup>**[src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs, line 334 [r1]](https://reviewable.io:443/reviews/nuget/nuget3/29#-JpJ8xShsoWXnfnjGxoU)** ([raw file](https://github.com/nuget/nuget3/blob/cd2bb7165df47e8325ddbcc12f97bd5e0a53f8e4/src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs#L334)):</sup>
This seemed unrelated but I was getting nullrefs without changing it.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/nuget/nuget3/29)

pranavkm:
<sup>**[src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs, line 334 [r1]](https://reviewable.io:443/reviews/nuget/nuget3/29#-JpJ8xShsoWXnfnjGxoU)** ([raw file](https://github.com/nuget/nuget3/blob/cd2bb7165df47e8325ddbcc12f97bd5e0a53f8e4/src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs#L334)):</sup>
If you rebase on dev, this should be fixed.

---

Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/nuget/nuget3/29)

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -27,6 +29,9 @@ namespace NuGet.Commands
 {
     public class RestoreCommand
     {
+        // Temporary until we have RESX and loc
```
emgarten: Do we plan to loc this for RTM? the deadline for locking down strings is coming up.

analogrelay: There are a bunch of strings in our trace output we'll also need to consider, especially if they end up in the Output Window...

yishaigalatzer:
String freeze is next week. So we need to do a pass and insert all the resx strings asap

analogrelay:
I'll rebase and merge this once I've got sign-off and work on resxifying ASAP after that. We need to lock down exactly what we want to resxify. For example:
- Exception Messages - yes, obviously
- Comments, Errors, Warnings, etc. written to build files (*.targets, etc.) - yes
- Logging output - Maybe? It changes rapidly and may have to change post string-freeze, but mostly for internal instrumentation purposes. However, it is ALSO written to the Output Window during a restore...

analogrelay:
I've added a resx string for this. Still need to make a pass and do the log messages I guess.

:watch:

PR #32, opened 2015-05-18
Dev johtaylo/unlisted

johnataylor:
- adding unlisted packages to api-v3 gather
- removed excessive unnecessary http calls from api-v3 metadata fetch
- adding unlisted to resolver logic (ordering in ResolverComparer)
- unit tests for resolver sort and resolve


yishaigalatzer:
@johnataylor does this PR needs to stay open?

PR #35, opened 2015-05-28
RestoreCommand nearest framework support

### LINKED ISSUE #611: Newtonsoft.Json 7.0.1-beta3 does not show up as a lock file target
1. install newtonsoft.json 7.0.1-beta3
2. build

newtonsoft.json.dll does not show up in the lock file under targets.

emgarten:
Fixes: https://github.com/NuGet/Home/issues/611

Adding CompareTest to ContentPropertyDefinition to allow finding the nearest framework in scenarios where multiple frameworks are compatible with a project framework, and neither can be ruled out by checking compatibility between the framework options.

### DIFF
--- a/src/NuGet.ContentModel/ContentPropertyDefinition.cs
+++ b/src/NuGet.ContentModel/ContentPropertyDefinition.cs
@@ -15,52 +15,61 @@ namespace NuGet.ContentModel
     public class ContentPropertyDefinition
     {
         public ContentPropertyDefinition(string name)
-            : this(name, null, null, null, null, false)
+            : this(name, null, null, null, null, null, false)
         {
         }
 
         public ContentPropertyDefinition(string name, IDictionary<string, object> table)
-            : this(name, table, null, null, null, false)
+            : this(name, table, null, null, null, null, false)
         {
         }
 
         public ContentPropertyDefinition(string name, Func<string, object> parser)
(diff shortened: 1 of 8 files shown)


emgarten:
//cc @anurse

src/NuGet.ContentModel/ContentPropertyDefinition.cs
```diff
@@ -146,6 +157,11 @@ public virtual bool TryLookup(string name, out object value)
 
         public Func<object, object, bool> CompatibilityTest { get; }
 
+        /// <summary>
+        /// Find the nearest compatible candidate.
+        /// </summary>
+        public Func<object, object, object, int> CompareTest { get; }
```
yishaigalatzer, line 163: this looks like a bug farm waiting to happen, and so do all the method calls with a million nulls. Can we track cleanup for U1

yishaigalatzer:
:shipit:

analogrelay:
Looks good to me, :shipit:. @davidfowl , might be good for you to take a quick look too.

src/NuGet.ContentModel/ContentPropertyDefinition.cs
emgarten, line 163: https://github.com/NuGet/Home/issues/684

PR #36, opened 2015-06-01
Renaming Core50 to netportable50

emgarten:
This change renames core50 to netportable50 and adds additional handling around portable frameworks to denote the difference between regular PCLs and .NETPortable 5.0. 

The order of compatibility is: 
UAP -> netcore50 -> netportable50

Note: NETPortable already existed as a framework synonym of .NETPortable in both NuGet.Frameworks and NuGet.Core, so only the framework version number can be used to differentiate between legacy PCLs and Core CLR portable libraries.

I've also added OriginalString to NuGetFramework to improve compatibility with unknown frameworks as part of this change.


emgarten:
//cc @anurse @pranavkm

PR #38, opened 2015-06-04
Normalize versions in package paths

emgarten:
This change updates DefaultPackagePathResolver to use normalized versions in paths.

For V3 we should avoid the problems non-normalized versions create when trying to find package paths.


emgarten:
//cc @pranavkm  @anurse

yishaigalatzer:
Does this change have no impact on packages.config scenarios?

emgarten:
packages.config uses a different package path resolver. We may want to rename this from "default" to V3 or something else, I think that it just kept the original name from the prototype branch.

DefaultPackagePathResolver reads the v3 folder structure for the restore command.

yishaigalatzer:
Yeah rename sounds good

yishaigalatzer:
:shipit: after rename is done

PR #39, opened 2015-06-04
Use Name instead of Path when setting up root ExternalProjectReference.

pranavkm:
cc @emgarten @anurse

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -59,7 +59,7 @@ namespace NuGet.Commands
             if (request.ExternalProjects.Any())
             {
                 externalProjectReference = new ExternalProjectReference(
-                    request.Project.FilePath,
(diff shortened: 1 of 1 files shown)


yishaigalatzer:
:shipit:

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -59,7 +59,7 @@ public RestoreCommand(ILogger logger)
             if (request.ExternalProjects.Any())
             {
                 externalProjectReference = new ExternalProjectReference(
-                    request.Project.FilePath,
+                    request.Project.Name,
```
pranavkm, line 62: `Project.Name` == `Project.FilePath` when we're dealing with NuGet.CommandLine, however they are different (Name maps to UniqueName) when dealing with VS build integrated project.

emgarten:
:shipit:

PR #40, opened 2015-06-04
Add PCL guardrails

analogrelay:
Adds PCL guardrails. Note: **currently does NOT support reading the `supports` data from the project.json**

To test it, go to `misc\RestoreTests\SupportsTest` and run `..\..\..\nuget3 restore`. You should see the `dnxcore50` restore work and the `sl40` restore fail. There is a partial example of PCL guardrails (refs without matching libs, etc.) in `misc\RestoreTests\PclExample` but it requires packages that contain `ref/dotnet` folders and such. To test a project against a "supports profile" (i.e. a TxM, RID pair you want to test restore against) you must specify it in the `--supports txm~rid` switch to nuget3 restore. For example `nuget3 restore --source <source of CLR packages> --supports netcore50~win8-aot`.

I will be adding the project.json integration shortly but I wanted to get the algorithm testable first :)

/cc @emgarten @lodejard @davidfowl @yishaigalatzer  @ericstj

### DIFF
--- a/misc/RestoreTests/PclExample/project.json
+++ b/misc/RestoreTests/PclExample/project.json
@@ -1,0 +1,13 @@
+{
+    "dependencies": {
+        "System.Linq": "4.0.0-beta-*",
+        "Microsoft.NETCore.Targets": "1.0.0-beta-*"
+    },
+    "frameworks": {
+        "core50": {}
+    },
+    "supports": {
+        "netcore50": ["win8-x86", "win8-x64"],
+        "dnxcore50": []
+    }
+}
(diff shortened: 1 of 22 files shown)


emgarten:
How about some unit tests for this?

analogrelay:
Yep, that's the plan. Just trying to get stuff up as quickly as possible :)

analogrelay:
Added a test for the scenario of a package that is incompatible with the framework. Still working on finding a good test case (short of writing my own package or making the tests depend upon internal builds) for the 'ref/' and 'lib/' matching.

And by added a test, I mean added asserts to an existing test that happened to test the exact scenario already :)

PR #45, opened 2015-06-09
Cache directory scans from PackageFolderReader.

pranavkm:


pranavkm:
cc @emgarten \ @yishaigalatzer 

Meant to reduce the number of disk scans once we resolve a package.

PR #49, opened 2015-06-11
Making the package manager detect and report errors for source paths …

harshgMSFT:
…which do not exist. Fixs #594.

### DIFF
--- a/src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
+++ b/src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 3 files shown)


harshgMSFT:
@yishaigalatzer @pranavkm @danliu

src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
```diff
@@ -43,6 +44,13 @@ public UISearchResourceV2(IPackageRepository repo)
         {
             return await Task.Run(() =>
             {
+                // Check if source is available.
```
yishaigalatzer, line 47: This comment should say: Check if the source is valid

src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
```diff
@@ -43,6 +44,13 @@ public UISearchResourceV2(IPackageRepository repo)
...
+                if (!IsHttpSource(V2Client.Source) && !IsLocalOrUNC(V2Client.Source))
+                {
+                    throw new InvalidOperationException(
```
yishaigalatzer, line 50: I assume you verified we don't crash here, and we show a proper error message

yishaigalatzer:
:shipit:

src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
harshgMSFT, line 50: yes and yes.

src/NuGet.Protocol.VisualStudio/UISearchResourceV2.cs
harshgMSFT, line 47: the source might be valid, but not available (network issues etc), .. if you think valid is better ill change that.

yishaigalatzer, line 47: The comment reads as applying only to the if. You might want to clarify.

I'm ok either way 

-----Original Message-----

PR #50, opened 2015-06-12
Switching to using Semaphore without timeouts in ConcurrencyUtilities.

pranavkm:
Ported from https://github.com/aspnet/dnx/commit/955fc0904614f3ca70bffaa65d7f134372497857

cc @deepakaravindr @emgarten @yishaigalatzer

### DIFF
--- a/src/NuGet.Common/ConcurrencyUtilities.cs
+++ b/src/NuGet.Common/ConcurrencyUtilities.cs
@@ -10,41 +10,35 @@ namespace NuGet.Common
 {
     internal static class ConcurrencyUtilities
     {
-        internal static async Task<T> ExecuteWithFileLocked<T>(string filePath, Func<Task<T>> action)
+        public static Task<TVal> ExecuteWithFileLocked<TVal>(string filePath, TimeSpan timeout, Func<bool, Task<TVal>> action)
         {
-            T result = default(T);
-
-            using (var filelock = new Mutex(initiallyOwned: false, name: FilePathToLockName(filePath)))
+            for (var i = 0; i < 3; ++i)
             {
-                while (true)
+                var createdNew = false;
+                var fileLock = new Semaphore(initialCount: 0, maximumCount: 1, name: FilePathToLockName(filePath),
+                    createdNew: out createdNew);
+                try
                 {
-                    try
+                    // If this lock is already acquired by another process, wait until we can acquire it
+                    if (!createdNew)
                     {
-                        if (filelock.WaitOne(1000))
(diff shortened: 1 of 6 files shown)


src/NuGet.Common/ConcurrencyUtilities.cs
```diff
@@ -10,41 +10,35 @@ namespace NuGet.Common
...
+                var fileLock = new Semaphore(initialCount: 0, maximumCount: 1, name: FilePathToLockName(filePath),
+                    createdNew: out createdNew);
+                try
                 {
-                    try
+                    // If this lock is already acquired by another process, wait until we can acquire it
+                    if (!createdNew)
                     {
-                        if (filelock.WaitOne(1000))
+                        var signaled = fileLock.WaitOne(timeout);
```
yishaigalatzer, line 25: you do need a much shorter timeout, because semaphores are not signalled when another process is killed, and then the retry of 3 is just wrong

pranavkm:
Discussed wirh @yishaigalatzer. @deepakaravindr will look at this since it needs more extensive changes.

src/NuGet.Common/ConcurrencyUtilities.cs
```diff
@@ -10,41 +10,35 @@ namespace NuGet.Common
 {
     internal static class ConcurrencyUtilities
     {
-        internal static async Task<T> ExecuteWithFileLocked<T>(string filePath, Func<Task<T>> action)
+        public static Task<TVal> ExecuteWithFileLocked<TVal>(string filePath, TimeSpan timeout, Func<bool, Task<TVal>> action)
```
deepakaravindr, line 13: You brought back the bool in the Func. Is it really needed?! Because, none of the callers are using it now

src/NuGet.Common/ConcurrencyUtilities.cs
```diff
@@ -10,41 +10,35 @@ namespace NuGet.Common
...
         {
-            T result = default(T);
-
-            using (var filelock = new Mutex(initiallyOwned: false, name: FilePathToLockName(filePath)))
+            for (var i = 0; i < 3; ++i)
             {
-                while (true)
+                var createdNew = false;
+                var fileLock = new Semaphore(initialCount: 0, maximumCount: 1, name: FilePathToLockName(filePath),
+                    createdNew: out createdNew);
```
deepakaravindr, line 19: Break the line. 1 line for each parameter

pranavkm, line 19: We're not taking this change.

PR #52, opened 2015-06-15
Remove colors from output

### LINKED ISSUE #746: RestoreCommand logs color information as text
The restore command writes out the literal string for colors and bolding to the log. This data should be removed when logging to the VS output window for UAP package installs/restores.

analogrelay:
Temporarily removing colors from the output as a tactical fix to get in to the UWP NuGet release. We should (and will) totally add them back (I'll make a bug for that) but we need to re-add them with code to disable them in VS, since the output window there doesn't support colors and it shows the literal ANSI escape codes.

/cc @emgarten @yishaigalatzer 

Fixes NuGet/Home#746


analogrelay:
Filed NuGet/Home#786 for restoring colors

emgarten:
:shipit:

PR #53, opened 2015-06-15
fix NuGet/Home#752 by allowing asm.ni.dll to match a ref assembly for asm

### LINKED ISSUE #752: [NuGet v3] Compatibility Checker should consider 'asm.ni.dll' a valid match for 'asm.dll'
When scanning for `lib` assemblies that match a given `ref` assembly, we should consider `lib/[txm]/foo.ni.dll` a valid match for `ref/[txm]/foo.dll`.

/cc @ericstj @davidfowl

analogrelay:
Fixes NuGet/Home#752

CLR packages will have some cases where there is a `ref/Foo.dll` and no `lib/Foo.dll`, instead there will be a `lib/Foo.ni.dll` (often inside a runtimes directory). This should be considered a match when pairing up ref and lib assemblies. We already put `.ni.dll` files into the lock file so this is a pretty minor and low-impact fix.

/cc @ericstj @emgarten @yishaigalatzer @davidfowl

### DIFF
--- a/misc/RestoreTests/NativeImageMatch/project.json
+++ b/misc/RestoreTests/NativeImageMatch/project.json
@@ -1,0 +1,14 @@
+{
+  "dependencies": {
+    "Microsoft.NETCore.UniversalWindowsPlatform": "1.0.0-beta-*"
+  },
+  "frameworks": {
+    "netcore50": {}
+  },
+  "runtimes": {
+    "win10-x86": {},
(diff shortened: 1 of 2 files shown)


analogrelay:
There's a minor issue, fix coming.

analogrelay:
Fixed, and verified against a sample provided by @ericstj.

ericstj:
:+1: LGTM

PR #54, opened 2015-06-15
fix NuGet/Home#784 by copying from remote provider directly to destination file

analogrelay:
Fixes NuGet/Home#784

/cc @emgarten @yishaigalatzer @davidfowl


yishaigalatzer:
I thought you mentioned ifdefs for corclr, but for now this looks good :shipit:

analogrelay:
Well, it turns out we were just loading the memory stream and dumping it straight into the file, then re-reading the file for the SHA, so the memory stream was useless.

emgarten:
:shipit:

PR #56, opened 2015-06-17
Adding delete On restart manager nuget3 changes

harshgMSFT:
@deepakaravindr @yishaigalatzer

### DIFF
--- a/src/NuGet.Packaging/PackageExtraction/PackagePathHelper.cs
+++ b/src/NuGet.Packaging/PackageExtraction/PackagePathHelper.cs
@@ -69,7 +69,7 @@ namespace NuGet.Packaging
                 filter.EndsWith(PackagingCoreConstants.NuspecExtension, StringComparison.OrdinalIgnoreCase));
 
             // Check for package files one level deep. We use this at package install time
-            // to determine the set of installed packages. Installed packages are copied to 
+            // to determine the set of installed packages. Installed packages are copied to
             // {id}.{version}\{packagefile}.{extension}.
             foreach (var dir in GetDirectories(root, String.Empty))
             {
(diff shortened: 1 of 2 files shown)


src/NuGet.Packaging/PackageExtraction/PackagePathHelper.cs
```diff
@@ -170,6 +170,25 @@ public static string GetInstalledPackageFilePath(PackageIdentity packageIdentity
...
+        public static string GetInstalledPackageFolder(PackageIdentity packageIdentity, PackagePathResolver packagePathResolver)
+        {
+            var packageLookupPaths = GetPackageLookupPaths(packageIdentity, packagePathResolver);
+
+            // TODO: Not handling nuspec-only scenarios
+            foreach (var packageLookupPath in packageLookupPaths)
+            {
+                if (Directory.Exists(packageLookupPath) ||
+                    (packageLookupPath.EndsWith(PackagingCoreConstants.NupkgExtension, StringComparison.OrdinalIgnoreCase) &&
+                    File.Exists(packageLookupPath)))
```
deepakaravindr, line 183: If packageLookupPath can be a filename, shouldn't you have to return Path.GetDirectoryName(packageLookupPath)? I will simply split the if into 2 and in the case of a file, return Path.GetDirectoryName

src/NuGet.Packaging/PackagePathResolver.cs
```diff
@@ -67,5 +67,10 @@ public virtual string GetInstalledPackageFilePath(PackageIdentity packageIdentit
         {
             return PackagePathHelper.GetInstalledPackageFilePath(packageIdentity, this);
         }
+
+        public virtual string GetInstalledPackageFolder(PackageIdentity packageIdentity)
```
deepakaravindr, line 71: Perhaps, add a (verbose) comment here, stating, how this differs from GetInstalledPath

deepakaravindr:
:shipit:

PR #58, opened 2015-06-18
fix NuGet/Home#818 by writing version range correctly

analogrelay:
/cc @davidfowl @emgarten @yishaigalatzer @pranavkm


src/NuGet.ProjectModel/LockFileFormat.cs
```diff
@@ -82,8 +103,10 @@ public LockFile Read(TextReader reader)
                     return ReadLockFile(token as JObject);
                 }
             }
-            catch
+            catch (Exception ex)
             {
+                log.LogError(Strings.FormatLog_ErrorReadingLockFile(ex.Message));
```
analogrelay, line 108: This will make it much easier to debug these in future :)

src/NuGet.ProjectModel/LockFileFormat.cs
```diff
@@ -177,7 +214,7 @@ private JProperty WriteTarget(LockFileTarget target)
             return new JProperty(key, json);
         }
 
-        private LockFileTarget ReadTarget(string property, JToken json)
+        private static LockFileTarget ReadTarget(string property, JToken json)
```
analogrelay, line 217: I started `static`ifying `LockFileFormat` but since the VS extension depends on it, I didn't want to make the final change until @emgarten was aware and onboard :)

PR #59, opened 2015-06-20
Adding support for fallback sources

pranavkm:
cc @anurse  @emgarten


analogrelay:
:shipit: - ping me when you merge this. I want to make a change to the arguments list as well (removing `--supports` which was only really there for testing)

PR #62, opened 2015-06-23
fix NuGet/Home#831 by skipping the compat check when there are unresolved packages

### LINKED ISSUE #831: Missing packages show guard rails errors if they are not found
1. Install newtonsoft.json -version 9.0.0  (does not exist)
2. Build

An error message saying Unable to resolve package will be displayed, but afterwards compatibility checks still run and give a generic message saying:
"Some packages are not compatible with .NETCore,Version=v5.0."

The error is misleading and a different error show be shown if an earlier restore caused the issue. Or we should error out and stop the restore after the first error when the package was determined to be missing.

analogrelay:
We now skip the compat check if there are unresolved packages, since the results will be invalid. This reduces error spew

Fixes NuGet/Home#831

/cc @emgarten @yishaigalatzer

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -76,22 +76,25 @@ namespace NuGet.Commands
                 lockFile.IsLocked = relockFile;
             }
 
-            // Scan every graph for compatibility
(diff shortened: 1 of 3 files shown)


yishaigalatzer:
@anurse should this PR be closed?

analogrelay:
yes

PR #63, opened 2015-06-23
fix NuGet/Home#838 by adding a Commit behavior

analogrelay:
Callers to restore now have to call `Commit` on the result to save the Lock File and MSBuild files to disk. This makes things a LOT easier for VS.

NOTE: The packages will be ALWAYS be downloaded, regardless of when `Commit` is called. Since the lock file is the thing that actually "enables" the newly downloaded packages, this seems fine.

/cc @davidfowl @emgarten @pranavkm @yishaigalatzer 

@emgarten This will require changes to your code, but I hope they will be fairly straightforward and mostly just removing code :P

Fixes NuGet/Home#838

### DIFF
--- a/src/NuGet.Commands/MSBuildRestoreResult.cs
+++ b/src/NuGet.Commands/MSBuildRestoreResult.cs
@@ -1,0 +1,150 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using System.Xml.Linq;
+using NuGet.Logging;
+
+namespace NuGet.Commands
+{
+    public class MSBuildRestoreResult
+    {
+        /// <summary>
+        /// Gets a boolean indicating if the necessary MSBuild file could be generated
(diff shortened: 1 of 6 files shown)


emgarten:
Much cleaner :)  :shipit: 

I'll update PackageManagement to work with the API changes.

src/NuGet.Commands/MSBuildRestoreResult.cs
```diff
@@ -0,0 +1,143 @@
+﻿using System;
```
pranavkm: License header.

src/NuGet.Commands/MSBuildRestoreResult.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            Success = false;
+            ProjectName = projectName;
+            ProjectDirectory = projectDirectory;
+            RepositoryRoot = string.Empty;
+            Props = Enumerable.Empty<string>();
+            Targets = Enumerable.Empty<string>();
+        }
+
+        public MSBuildRestoreResult(string projectName, string projectDirectory, string repositoryRoot, IEnumerable<string> props, IEnumerable<string> targets)
```
pranavkm: Break this itno smaller lines.

pranavkm:
:shipit:

PR #64, opened 2015-06-23
fix NuGet/Home#716 by adding a warning when a dependency is bumped up

analogrelay:
Fixes NuGet/Home#716

Simple warning that just tells the user if a dependency is bumped up from what was specified in the original project.json.

/cc @emgarten @yishaigalatzer


analogrelay:
Ping, updated.

emgarten:
:shipit:

PR #65, opened 2015-06-23
fix NuGet/Home#839 by RESXing ALL THE THINGS

analogrelay:
Fixes NuGet/Home#839

Also removes some diagnostic/prototype stuff.

/cc @emgarten @yishaigalatzer


yishaigalatzer:
Some of the strings are missing dots at the end of the sentence, other than that :shipit:

PR #66, opened 2015-06-25
Fixes https://github.com/NuGet/Home/issues/846.

deepakaravindr:
Added a CancellationToken to RestoreCommand

@yishaigalatzer , @emgarten , @pranavkm

### DIFF
--- a/src/NuGet.DependencyResolver/SourceRepositoryDependencyProvider.cs
+++ b/src/NuGet.DependencyResolver/SourceRepositoryDependencyProvider.cs
@@ -76,7 +76,11 @@ namespace NuGet.DependencyResolver
 
             using (var nupkgStream = await _findPackagesByIdResource.GetNupkgStreamAsync(identity.Name, identity.Version, cancellationToken))
             {
-                await nupkgStream.CopyToAsync(stream, bufferSize: 8192, cancellationToken: cancellationToken);
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // If the stream is already available, do not stop in the middle of copying the stream
+                // Pass in CancellationToken.None
+                await nupkgStream.CopyToAsync(stream, bufferSize: 8192, cancellationToken: CancellationToken.None);
             }
         }
 
--- a/src/NuGet.Packaging/PackageExtraction/NuGetPackageUtils.cs
+++ b/src/NuGet.Packaging/PackageExtraction/NuGetPackageUtils.cs
@@ -47,6 +47,12 @@ namespace NuGet.Packaging
                         log.LogInformation($"Installing {packageIdentity.Id} {packageIdentity.Version}");
 
                         Directory.CreateDirectory(targetPath);
+
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        // Do not stop the package extraction after this point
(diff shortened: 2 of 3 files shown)


pranavkm:
Could you make this change as two separate commits? One with the CancellationToken and with the cleanup? This way it's easy to backout \ fix one if it's problematic.

src/NuGet.DependencyResolver/SourceRepositoryDependencyProvider.cs
```diff
@@ -76,7 +76,11 @@ public class SourceRepositoryDependencyProvider : IRemoteDependencyProvider
 
             using (var nupkgStream = await _findPackagesByIdResource.GetNupkgStreamAsync(identity.Name, identity.Version, cancellationToken))
             {
-                await nupkgStream.CopyToAsync(stream, bufferSize: 8192, cancellationToken: cancellationToken);
+                cancellationToken.ThrowIfCancellationRequested();
+
+                // If the stream is already available, do not stop in the middle of copying the stream
```
pranavkm, line 81: Why are we doing this?

deepakaravindr, line 81: This is the last point at which we can cancel the restore safely. Do you ask this, because, you consider that package download to be the most expensive part of the restore operation?

If you are asking about passing CancellationToken.None to Stream.CopyToAsync,
I don't want to pass in the real CancellationToken to Stream.CopyToAsync. I think it will stop the copy abruptly if the CancellationToken was cancelled. Previously, this was never an issue, because, well, we were not passing the real CancellationToken

deepakaravindr:
@pranavkm , Your request to split the commits is generally valid. But, in this case, the real change is just the addition of the CancellationToken parameter. So, I don't think it matters that much

src/NuGet.Packaging/PackageExtraction/NuGetPackageUtils.cs
```diff
@@ -47,6 +47,12 @@ public static class NuGetPackageUtils
                         log.LogInformation($"Installing {packageIdentity.Id} {packageIdentity.Version}");
```
yishaigalatzer, line 47: This needs to become a resx string

src/NuGet.Packaging/PackageExtraction/NuGetPackageUtils.cs
```diff
@@ -47,6 +47,12 @@ public static class NuGetPackageUtils
...
 
                         Directory.CreateDirectory(targetPath);
+
+                        cancellationToken.ThrowIfCancellationRequested();
+
+                        // Do not stop the package extraction after this point
+                        // based on CancellationToken
```
yishaigalatzer, line 54: move the throw above the directory creation

yishaigalatzer:
:shipit: in two commits after review comments addressed
1. Cleanup
2. Code changes

deepakaravindr:
Done. Fixed with https://github.com/NuGet/NuGet3/commit/ed45428d130935fee2e95f71f737bcf7c5acbc83 and https://github.com/NuGet/NuGet3/commit/6d01de61db6b0e7e164e9c042d5be005b983c9a2

PR #67, opened 2015-06-29
RestoreCommand performance improvements

emgarten:
RestoreCommand performance improvements.
1. Reuse libraries from the previous lock file if the SHA is the same
2. Read the SHA512 from disk instead of hashing the nupkg
3. Read the nuspec from disk instead of from the nupkg during library target entry creation.
4. Download all nupkgs even if they are not used in the graph. This helps keep us from having to go online during the next lock file creation to find them again.
5. Cache runtime graphs
6. Store library file lists to avoid fetching them from disk multiple times. If the previous library can be reused the files are used from that.

//cc @davidfowl @anurse @yishaigalatzer

### DIFF
--- a/src/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
+++ b/src/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
@@ -1,7 +1,11 @@
 // Copyright (c) .NET Foundation. All rights reserved.
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
+using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Threading.Tasks;
+using NuGet.LibraryModel;
+using NuGet.Packaging.Core;
 
 namespace NuGet.DependencyResolver
 {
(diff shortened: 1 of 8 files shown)


src/NuGet.DependencyResolver.Core/Remote/RemoteWalkContext.cs
```diff
@@ -12,10 +16,23 @@ public RemoteWalkContext()
             ProjectLibraryProviders = new List<IRemoteDependencyProvider>();
             LocalLibraryProviders = new List<IRemoteDependencyProvider>();
             RemoteLibraryProviders = new List<IRemoteDependencyProvider>();
+
+            FindLibraryEntryCache = new ConcurrentDictionary<LibraryRangeCacheKey, Task<GraphItem<RemoteResolveResult>>>();
```
davidfowl: Move this to another context.

analogrelay:
:shipit: Awesome! I wanted to get an optimization pass in but didn't end up with time so I'm glad you got a chance to dive in and do it.

PR #68, opened 2015-06-30
DotNet framework compatibility

emgarten:
This change adds support for installing packages with 'dotnet' libraries into projects with known TxMs except for a set of  explicitly incompatible ones such as net40, sl5, wp7, native.

https://github.com/NuGet/Home/issues/768

//cc @davidfowl @anurse @yishaigalatzer

### DIFF
--- a/src/NuGet.Frameworks/DefaultFrameworkMappings.cs
+++ b/src/NuGet.Frameworks/DefaultFrameworkMappings.cs
@@ -332,11 +332,68 @@ namespace NuGet.Frameworks
                                 new FrameworkRange(
                                     new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.WinRT, FrameworkConstants.EmptyVersion),
                                     new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.WinRT, new Version(4, 5, 0, 0)))),
+
+                            // Dotnet compatibility white list
+                            // net45 and up support dotnet
+                            new OneWayCompatibilityMappingEntry(
+                                new FrameworkRange(
+                                    new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.Net, new Version(4, 5, 0, 0)),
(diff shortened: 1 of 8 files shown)


src/NuGet.Frameworks/DefaultFrameworkMappings.cs
```diff
@@ -332,6 +332,54 @@ public IEnumerable<OneWayCompatibilityMappingEntry> CompatibilityMappings
...
+                                    includeMin: false,
+                                    includeMax: true),
+                                FrameworkConstants.DotNetAll),
+
+                            // wp8 and up support dotnet
+                            new OneWayCompatibilityMappingEntry(
+                                new FrameworkRange(
+                                    new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.WindowsPhone, new Version(8, 0, 0, 0)),
+                                    new NuGetFramework(FrameworkConstants.FrameworkIdentifiers.WindowsPhone, FrameworkConstants.MaxVersion)),
+                                FrameworkConstants.DotNetAll),
```
analogrelay, line 365: Let's make this `dotnet50` for now. Things are still up-in-the-air a bit and it'd be better to be conservative for now.

test/NuGet.Frameworks.Test/CompatibilityTests.cs
```diff
@@ -237,13 +249,37 @@ public void Compatibility_DotNetNeg(string framework)
...
+        [InlineData("monoandroid40")]
+        [InlineData("monomac")]
+        [InlineData("xamarinios")]
+        [InlineData("xamarinmac")]
+        [InlineData("xamarinpsthree")]
+        [InlineData("xamarinpsfour")]
+        [InlineData("xamarinpsvita")]
+        [InlineData("xamarinxboxthreesixty")]
+        [InlineData("xamarinxboxone")]
+        [InlineData("sl6")]
```
analogrelay, line 272: I think we want to block `sl` from `dotnet` entirely, rather than just up to `sl5`

analogrelay:
Minor suggestions but other than those, :shipit:

PR #69, opened 2015-06-30
update lock file version to 1

analogrelay:
It's a complicated one, I know :P.

/cc @davidfowl @lodejard @yishaigalatzer @emgarten @jasonmalinowski


yishaigalatzer:
:shipit: but lets give shiproom sometime before actually committing. I'm assuming its fine, but we can wait till 7pm or so to push

PR #70, opened 2015-07-01
Introduce a targeted method to perform v2 -> v3 migrations.

pranavkm:
Fixes https://github.com/NuGet/Home/issues/861

cc @deepakaravindr @zhili1208 @yishaigalatzer

### DIFF
--- a/src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
+++ b/src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
@@ -171,30 +171,39 @@ namespace NuGet.Configuration
             return loadedPackageSources;
         }
 
-        public static void MigrateDefaultFeedToNewerProtocol()
+        public static void AddDefaultSourcesToRootNuGetConfig()
         {
             // Load the root settings (%AppData% one)
             var settings = Configuration.Settings.LoadDefaultSettings(root: null, configFileName: null, machineWideSettings: null);
-            if (settings != null)
-            {
-                MigrateDefaultFeedToNewerProtocolCore(settings);
-
-            }
+            AddDefaultSourcesToRootNuGetConfigCore(settings);
         }
 
         // Internal for unit testing.
-        internal static void MigrateDefaultFeedToNewerProtocolCore(ISettings settings)
+        internal static void AddDefaultSourcesToRootNuGetConfigCore(ISettings settings)
         {
             var sourceSettings = settings.GetSettingValues(ConfigurationContants.PackageSources).ToList();
 
             var settingToMigrate = sourceSettings
                     .Where(f => f.Key.Equals(NuGetConstants.FeedName, StringComparison.OrdinalIgnoreCase))
                     .ToArray();
+            var v3Setting = new SettingValue(
+                NuGetConstants.FeedName,
+                NuGetConstants.V3FeedUrl,
+                isMachineWide: false,
+                priority: 1);
+            v3Setting.AdditionalData[ConfigurationContants.ProtocolVersionAttribute] = "3";
+
+            var v2Setting = new SettingValue(
+               NuGetConstants.FeedName,
+               NuGetConstants.V2FeedUrl,
+               isMachineWide: false,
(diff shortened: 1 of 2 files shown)


zhili1208:
For me, this looks like a new feature, not a fix.
Consider following case:
on a clean machine, there is no NuGet.Config. 
User Open Visual Studio and NuGet Powershell.
NuGet initialized and generated NuGet.Config in AppData..
Expect: NuGet.Config should be like there are two nuget.org sources(v2 and v3) and machine wide source is disabled.
actual: there are no package sources in NuGet.Config, machine wide source is disabled.
this also will cause another issue, if user open a solution and there is a nuget.config (contains one package source)in that solution. then user just open setting option , disable local source and save.
the two nuget.org sources will go to local NuGet.Config.

For another issue, relative path will be replace by full path. this does not fix that.
Consider following case:
Open a solution which contain a local nuget.config. The path in local nuget.config is relative path.
disable local package source and save.
the relative path in local nuget,config will become full path.

pranavkm:
This doesn't touch the local NuGet.config - see https://github.com/NuGet/NuGet3/pull/70/files#diff-a8ecdfdb7a533f7c7b9c7aea4b58bf77R177. 

The one about missing root NuGet.config makes sense - we could potentially create one if it doesn't already exist - but didn't that work before your code change in https://github.com/NuGet/NuGet.VisualStudioExtension/commit/e01289e3cea45db6a1dfdd05e0144fc8a9ae3a5f?

pranavkm:
The relative path one is problematic - but something we've always had (reproes with NuGet v 2.8.6) whenever `SavePackageSource` is invoked. That said, this change does avoid modifying relative paths in the NuGet.Config - I can add a test verifying that. Filed a bug for this - https://github.com/NuGet/Home/issues/865

zhili1208:
NO, this didn't work before my code change in.

pranavkm:
Updated

pranavkm:
BUMP @zhili1208 \  @deepakaravindr \ @yishaigalatzer

src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -171,13 +171,71 @@ public IEnumerable<PackageSource> LoadPackageSources()
...
+            }
+            else if (settingToMigrate.Any(setting => ReadProtocolVersion(setting) == 3))
+            {
+                // Source already exists for the newest protocol version. Do nothing.
+                return;
+            }
+            else
+            {
+                Debug.Assert(settingToMigrate.Length > 0);
+                var index = sourceSettings.IndexOf(settingToMigrate[0]);
```
deepakaravindr, line 216: At this point, you expect that there is only element in 'settingToMigrate'? And, that it is v2 'nuget.org'. If so, I would assert that

pranavkm, line 216: Sure.

src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -432,13 +490,13 @@ public void SavePackageSources(IEnumerable<PackageSource> sources)
             var existingSettings = (Settings.GetSettingValues(ConfigurationContants.PackageSources, isPath: true) ??
```
deepakaravindr, line 490: Don't you have to fix SavePackageSources method in any case. Otherwise, any changes being made to local nuget.config file, which is read-only, will be a problem, right?

With one of your changes, I believe, when you change a package source, it will change the nuget.config it comes from based on priority, instead of always writing to global nuget.config

So, in NuGet 3, it is more likely for the local nuget.config be updated compared to NuGet 2.8.6

If this cannot be fixed with this change, please file a bug to track this issue

pranavkm, line 490: Filed https://github.com/NuGet/Home/issues/868 to track this.

deepakaravindr:
:shipit:

PR #71, opened 2015-07-01
Re-run more of the restore logic for the compatibility check

davidfowl:
- Move core restore logic into a helper function
- Use it to restore and do the compat check
- Corrected a test

https://github.com/NuGet/Home/issues/852

/cc @yishaigalatzer @anurse @emgarten


emgarten:
:shipit:

analogrelay:
:shipit:

yishaigalatzer:
Can this PR be closed?

PR #74, opened 2015-07-01
Fix the formatting of lock file dependency groups

analogrelay:
Also fixes a bug where reference assemblies were being added to the original project dependencies when resolving dependencies for .NETFramework targets.

This is a fix @emgarten needs for his code to improve no-op restores.

/cc @davidfowl @yishaigalatzer

### DIFF
--- a/test/NuGet.Commands.Test/RestoreCommandTests.cs
+++ b/test/NuGet.Commands.Test/RestoreCommandTests.cs
@@ -423,7 +423,7 @@ namespace NuGet.Commands.Test
             const string project = @"
 {
     ""dependencies"": {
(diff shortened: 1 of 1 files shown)


src/NuGet.LibraryModel/LibraryRange.cs
```diff
@@ -28,6 +29,40 @@ public override string ToString()
             return output;
         }
 
+        public string ToLockFileDependencyGroupString()
```
analogrelay, line 32: Definitely taking suggestions on better names...

emgarten:
:shipit:

This goes with https://github.com/NuGet/Home/issues/844

PR #77, opened 2015-07-02
Do not add empty lines when attempting to preserve formatting that ap…cd 

pranavkm:
…pears prior to clear elements

### DIFF
--- a/src/NuGet.Configuration/Settings/Settings.cs
+++ b/src/NuGet.Configuration/Settings/Settings.cs
@@ -519,9 +519,15 @@ namespace NuGet.Configuration
                 }
             }
 
-            foreach (var element in nodesToRemove)
+            // Special case for the scenario where the clear element is the last element in the
+            // section (followed by whitespace and comments). In this case, we can avoid removing any
+            // node and preserving the original formatting.
+            if (nodesToRemove.Any(node => node.NodeType == XmlNodeType.Element))
             {
-                element.Remove();
+                foreach (var element in nodesToRemove)
+                {
+                    element.Remove();
+                }
             }
         }
 
--- a/src/NuGet.Configuration/Settings/Settings.cs
+++ b/src/NuGet.Configuration/Settings/Settings.cs
@@ -472,6 +472,8 @@ namespace NuGet.Configuration
                 sectionElement = GetOrCreateSection(ConfigXDocument.Root, section);
             }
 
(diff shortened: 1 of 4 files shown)


pranavkm:
cc @yishaigalatzer @zhili1208 @deepakaravindr

test/NuGet.Configuration.Test/PackageSourceProviderTests.cs
```diff
@@ -798,19 +796,17 @@ public void SavePackageSourcesWithOnlyClear()
...
 <configuration>
     <packageSources>
-        <clear /> 
+        <clear />
         <add key=""test3"" value=""https://test3.net"" />
     </packageSources>
     <disabledPackageSources>
-        <clear />
-    </disabledPackageSources>
+        <clear /></disabledPackageSources>
```
pranavkm: It ends up looking messed up in this scenario (the `clear` element is the last element) - I guess we can special case for this if it's pertinent?

src/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -496,20 +498,28 @@ private static void RemoveElementAfterClearTag(XElement sectionElement)
...
-            var removeElmentList = new List<XElement>();
-            foreach (var element in sectionElement.Elements())
+            var nodesToRemove = new List<XNode>();
+            foreach (var node in sectionElement.Nodes())
             {
+                if (node.NodeType != XmlNodeType.Element)
+                {
+                    nodesToRemove.Add(node);
+                    continue;
+                }
```
yishaigalatzer, line 508: nice use of continue

test/NuGet.Configuration.Test/PackageSourceProviderTests.cs
yishaigalatzer: that would be nice, but not critical

yishaigalatzer:
:shipit: after you make a decision about the </clear> as last element. I don't think its really common or ever a thing to have clear as last element, but this is the same behavior for sources we should fix

pranavkm:
Updated with a fix for `clear` being the last node.

pranavkm:
cc @danliu

pranavkm:
Merged in https://github.com/NuGet/NuGet3/commit/16327a708887aabdfc20d02631897a8a731ab1bf

PR #78, opened 2015-07-03
Avoid writing out the lock file for noop restores

emgarten:
This change allows RestoreResult.Commit(ILogger) to no-op if the lock file is identical to the existing lock file.

The majority of this change is adding Equals and GetHashCode methods to LockFile and all child types.

Fixes https://github.com/NuGet/Home/issues/844

//cc @anurse @yishaigalatzer @davidfowl @pranavkm


analogrelay:
:shipit:

PR #79, opened 2015-07-05
Several fixes to make cycle detection and downgrade detection work

davidfowl:
- Print out cycles and downgrades after conflict resolution
- Fix bug in runtime graph resolution where entries were being overwritten by merges


src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -93,31 +125,42 @@ public Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library, NuGe
...
+                foreach (var d in item.Data.Dependencies)
+                {
+                    if (d != dependency && d.Name == library.Name)
                     {
-                        return false;
+                        if (d.LibraryRange.VersionRange != null &&
+                            library.VersionRange != null &&
+                            d.LibraryRange.VersionRange.MinVersion < library.VersionRange.MinVersion)
+                        {
+                            return DependencyResult.PotentiallyDowngraded;
```
analogrelay, line 155: Just so I'm clear, if a `PotentiallyDowngraded` dependency is not actually downgraded, it's `Eclipsed`, right? Just confirming that we don't need to walk down the dependencies of a `PotentiallyDowngraded` node because no matter what, it will never be used in the final graph.

src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
```diff
@@ -50,37 +85,34 @@ public Task<GraphNode<RemoteResolveResult>> WalkAsync(LibraryRange library, NuGe
...
                         runtimeGraph,
-                        ChainPredicate(predicate, node.Item, dependency)));
-
-                    if (!string.IsNullOrEmpty(runtimeName)
-                        && runtimeGraph != null)
+                        ChainPredicate(predicate, node, dependency)));
+                }
+                else
+                {
+                    // Keep the node in the tree if we need to look at it later
```
analogrelay, line 105: A further comment here explaining that we don't need to walk down this part of the graph (see my below comment) would be useful.

src/NuGet.DependencyResolver.Core/Remote/RemoteDependencyWalker.cs
davidfowl, line 155: That's correct.

analogrelay:
Looks good, would like a few more small comments in the code so future generations don't have to figure this all out again :).

Other than that, :shipit:

PR #81, opened 2015-07-07
Remove cyclic nodes once they've been evaluated.

pranavkm:
Partial fix for https://github.com/NuGet/Home/issues/903

cc @anurse  \ @davidfowl

### DIFF
--- a/src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
+++ b/src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
@@ -45,6 +45,11 @@ namespace NuGet.DependencyResolver
                 if (node.Disposition == Disposition.Cycle)
                 {
                     cycles.Add(node);
+
+                    // Remove this node from the tree so the nothing else evaluates this.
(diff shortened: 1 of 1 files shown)


src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
```diff
@@ -45,6 +45,11 @@ private enum WalkState
                 if (node.Disposition == Disposition.Cycle)
                 {
                     cycles.Add(node);
+
+                    // Remove this node from the tree so the nothing else evaluates this.
+                    // This is ok since we have a parent pointer and we can still print the path
+                    node.OuterNode.InnerNodes.Remove(node);
```
yishaigalatzer, line 51: are all the "dotting" guaranteed to never return null?

pranavkm, line 51: Yup - a cyclic node always has a parent. cc @davidfowl

analogrelay:
:shipit:

PR #83, opened 2015-07-08
Prevent TryResolveConflict from throwing if a node without a version …

pranavkm:
…range

is specified.


pranavkm:
cc @yishaigalatzer

yishaigalatzer:
:shipit:

PR #86, opened 2015-07-14
Allow project.json package restore to work against authenticated feeds.

pranavkm:
Fixes https://github.com/NuGet/Home/issues/941

cc @emgarten @feiling @yishaigalatzer

### DIFF
--- a/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
@@ -82,7 +82,7 @@ namespace NuGet.Protocol.Core.v3
                 {
                     CredentialStore.Instance.Add(uri, credential);
                 }
-                return new ProxyWebRequestHandler()
+                return new CredentialPromptWebRequestHandler()
                 {
                     Proxy = proxy,
                     Credentials = credential
@@ -90,7 +90,7 @@ namespace NuGet.Protocol.Core.v3
             }
         }
 
-        private class ProxyWebRequestHandler : WebRequestHandler
+        private class CredentialPromptWebRequestHandler : WebRequestHandler
         {
             protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
             {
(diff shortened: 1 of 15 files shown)


src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
```diff
@@ -104,6 +104,20 @@ private class ProxyWebRequestHandler : WebRequestHandler
...
+                        {
+                            if (HttpHandlerResourceV3.PromptForProxyCredentials == null)
+                            {
+                                // Fail this operation.
+                                response.EnsureSuccessStatusCode();
+                            }
+                            else
+                            {
+                                // PromptForProxyCredentials is a misnomer. It can prompt for proxy credentials or request credentials.
+                                Credentials = HttpHandlerResourceV3.PromptForProxyCredentials(request.RequestUri, /* webProxy */ null);
```
yishaigalatzer: Should we rename to PromptForCredentials ?

yishaigalatzer:
Looks good to me, :shipit: when @feiling is happy and also signsoff on testing it

src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
pranavkm: Ideally it would be another delagate that didn't take the proxy parameter. But we could live with the rename.

feiling: I think it's much better to add a new delegate HttpHandlerResourceV3.PromptForCredentials(). PromptForProxyCredentails() does exactly what's it's named: asking for Proxy credentials. Thus, it also syncs with the v2 ProxyCache.

So similarly, the new PromptForCredentials() would need to sync with v2 CredentialStore.

you also need to add a new delegate, say, HttpHandlerResourceV3.CredentialPassed, which, similar to HttpHandlerResourceV3.ProxyPassed, syncs v3 credentials with v2.

feiling: Looks good for now. But we should clean this up later.

pranavkm:
Updated. I didn't notice this with the earlier change - but we were logging an error when attempting to set the credential on a the `CredentialPromptMessageHandler` instance the second time around. The way to address is to ensure we use a new Message handler for each request. The changes here are primarily around supporting this.

PR #88, opened 2015-07-15
Imports support for fallback frameworks

emgarten:
This change adds support for fallback frameworks under the "imports" property.

The additional framework is passed around a FallbackFramework which extends NuGetFramework. GetNearest checks for this type and handles it at the lowest level. This makes the change much less invasive than changing everything to take multiple frameworks.

Warnings are displayed if the fallback framework was used, this is done by building the target library using the normal fallback framework, and using the project framework only. If there are differences the warning is logged once per library.

``` json
                "frameworks": {
                    "dotnet": {
                        "imports": "portable-net452+win81",
                        "warn"": false
                    }
```

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -513,12 +513,18 @@ namespace NuGet.Commands
 
             var libraries = lockFile.Libraries.ToDictionary(lib => Tuple.Create(lib.Name, lib.Version));
 
+            var warnForImports = project.TargetFrameworks.Any(framework => framework.Warn);
+            var librariesWithWarnings = new HashSet<LibraryIdentity>();
+
             // Add the targets
             foreach (var targetGraph in targetGraphs)
             {
                 var target = new LockFileTarget();
                 target.TargetFramework = targetGraph.Framework;
                 target.RuntimeIdentifier = targetGraph.RuntimeIdentifier;
+
+                var fallbackFramework = target.TargetFramework as FallbackFramework;
+                var warnForImportsOnGraph = warnForImports && fallbackFramework != null;
 
                 foreach (var library in targetGraph.Flattened.Select(g => g.Key).OrderBy(x => x))
                 {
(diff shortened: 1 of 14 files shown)


emgarten:
//cc @yishaigalatzer @davidfowl @pranavkm @deepakaravindr

emgarten:
//cc @anurse would you take a look at this also?

src/NuGet.ProjectModel/JsonPackageSpecReader.cs
```diff
@@ -307,6 +319,40 @@ private static bool BuildTargetFrameworkNode(PackageSpec packageSpec, KeyValuePa
...
+
+            if (importsProperty != null)
+            {
+                var importFramework = NuGetFramework.Parse(importsProperty.ToString());
+
+                if (importFramework.IsPCL)
+                {
+                    // PCLs are the only frameworks allowed here, other values will be ignored
+                    framework = importFramework;
+                }
```
pranavkm, line 336: `else throw`?

src/NuGet.Frameworks/FrameworkReducer.cs
```diff
@@ -446,8 +463,8 @@ private bool IsBetterPCL(NuGetFramework current, NuGetFramework considering)
             }
 
             // Take the highest version of .NET if no winner could be determined, this is usually a good indicator of which is newer
-            var consideringNet = consideringFrameworks.Where(f => StringComparer.OrdinalIgnoreCase.Equals(f.Framework, FrameworkConstants.FrameworkIdentifiers.Net)).First();
-            var currentNet = currentFrameworks.Where(f => StringComparer.OrdinalIgnoreCase.Equals(f.Framework, FrameworkConstants.FrameworkIdentifiers.Net)).First();
+            var consideringNet = consideringFrameworks.Where(f => StringComparer.OrdinalIgnoreCase.Equals(f.Framework, FrameworkConstants.FrameworkIdentifiers.Net)).FirstOrDefault();
```
pranavkm, line 466: `var consideringNet = consideringFrameworks.FirstOrDefault(f => StringComparer.OrdinalIgnoreCase.Equals(f.Framework, FrameworkConstants.FrameworkIdentifiers.Net));`

src/NuGet.Frameworks/FallbackFramework.cs
```diff
@@ -0,0 +1,52 @@
...
+    {
+        /// <summary>
+        /// Secondary framework to fall back to.
+        /// </summary>
+        public NuGetFramework Fallback { get; }
+
+        public FallbackFramework(NuGetFramework framework, NuGetFramework fallbackFramework)
+            : base(framework)
+        {
+            Fallback = fallbackFramework;
```
pranavkm, line 18: Could you add a arg null check here for `fallbackFramework`? We should never be calling this with a null value so let's codify it.

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -538,6 +544,29 @@ private bool ResolutionSucceeded(IEnumerable<RestoreTargetGraph> graphs)
                         correctedPackageName: library.Name);
 
                     target.Libraries.Add(targetLibrary);
+
+                    // Log warnings if the target library used the fallback framework
+                    if (warnForImportsOnGraph && !librariesWithWarnings.Contains(library))
```
pranavkm, line 549: Wouldn't this warn if the target framework says don't warn i.e. one tfm says warn and the other doesn't?

pranavkm, line 549: Feel free to create a follow up item for this. Seem super low pri.

pranavkm:
:shipit:

src/NuGet.Commands/RestoreCommand.cs
emgarten, line 549: yes, it would, but currently there is only ever one framework for these build integrated projects.

src/NuGet.Frameworks/FallbackFramework.cs
emgarten, line 18: Good call, I'll update this.

src/NuGet.Commands/RestoreCommand.cs
pranavkm, line 549: > ever one framework for these build integrated projects.

So far :imp:

PR #90, opened 2015-07-17
Use NET45 instead of DNX451 directive

david-driscoll:
- NuGet.Protocol.Core.v3 is built for net45 not dnx451, so the code path was always looking for the Environment variable instead of the SpecialFolder.


emgarten:
@david-driscoll thanks for the fix!

NuGet/NuGet3@0358c0faf12274638e2312f47c236431305bd90b

PR #93, opened 2015-07-27
fixed https://github.com/NuGet/Home/issues/998

zhili1208:
the root casue of this bug is settings loaded same config file twice.
when user use nuget.exe sources add command to add source to local config file by using -ConfigFile,
and if config file name is NuGet.config, loadAppDataConfig and hierarchy Config search will add same config file to settings twice.(also exist in 2.8.6)
in this case, when we add a new source to the config file, for first setting, it will add source and save, for second setting, no new source and save. becasue they are same config file, so the second one overwrite the first one.

the fix is never add user speficific config to settings during hierarchy config search.

### DIFF
--- a/src/NuGet.Configuration/Settings/Settings.cs
+++ b/src/NuGet.Configuration/Settings/Settings.cs
@@ -138,9 +138,9 @@ namespace NuGet.Configuration
                     validSettingFiles.AddRange(
                         GetSettingsFileNames(root)
                             .Select(f => ReadSettings(root, f))
-                            .Where(f => f != null));
+                            .Where(f => f != null && !f.ConfigFilePath.Equals(Path.GetFullPath(Path.Combine(root, configFileName??"")))));
                 }
-
+               
                 if (loadAppDataSettings)
                 {
                     LoadUserSpecificSettings(validSettingFiles, root, configFileName, machineWideSettings);


zhili1208:
@deepakaravindr @pranavkm

pranavkm:
Did this not work in the past (2.8.6)?

pranavkm:
Shouldn't it not be walking up the hierarchy if a specific config file is specified? This fix looks quite wrong.

zhili1208:
@pranavkm Yes, this work in the past because we only add source to appdata config in the past. I'm not sure do we need to walk up the hirerarchy if a specific config file is specified. But in the past we did. I don't know how many people depends on this behavior, it might break people if we change this behavior

yishaigalatzer:
Please don't use the bug link in the title, put it in the comment so its clickable and I can read what this is about

yishaigalatzer:
@zhili1208 @pranavkm this issue is now stuck for almost 3 weeks. Can we get to a conclusion?

src/NuGet.Configuration/Settings/Settings.cs
```diff
@@ -138,9 +138,9 @@ public string ConfigFilePath
                     validSettingFiles.AddRange(
                         GetSettingsFileNames(root)
                             .Select(f => ReadSettings(root, f))
-                            .Where(f => f != null));
+                            .Where(f => f != null && !f.ConfigFilePath.Equals(Path.GetFullPath(Path.Combine(root, configFileName??"")))));
```
pranavkm, line 141: Fix formatting.

pranavkm:
:shipit:

zhili1208:
found another issue related to this, the scenario is when user have hierarchy config and use nuget.exe sources command add source to specified config file, the source will be added to upper hierarchy config file, will update this pr when I fix the issue.

zhili1208:
recall this PR

PR #98, opened 2015-08-05
Fix null reference exception from dependency conflict checks

emgarten:
During the graph walk a node may be a framework reference which does not contain a version. When checking for conflicts this throws a null ref. The fix for this is to filter based on type constraints to avoid these references.

https://github.com/NuGet/Home/issues/1116

//cc @pranavkm @yishaigalatzer @deepakaravindr


deepakaravindr:
:shipit:

PR #99, opened 2015-08-06
Daravind/issue 1062

deepakaravindr:
Fixes https://github.com/NuGet/Home/issues/1098
and adds test for
https://github.com/NuGet/Home/issues/1062

@yishaigalatzer @emgarten @pranavkm

### DIFF
--- a/src/NuGet.Configuration/Utility/SettingsUtility.cs
+++ b/src/NuGet.Configuration/Utility/SettingsUtility.cs
@@ -130,16 +130,16 @@ namespace NuGet.Configuration
                 throw new ArgumentNullException(nameof(settings));
             }
 
-            // Note that the value GlobalPackagesFolder must be a full path and not a relative path
-            // Read it like a string and explicitly not as a path
-            var path = settings.GetValue(ConfigSection, GlobalPackagesFolderKey, isPath: false);
+            var path = Environment.GetEnvironmentVariable(GlobalPackagesFolderEnvironmentKey);
             if (!string.IsNullOrEmpty(path))
             {
                 path = path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
                 return path;
             }
 
-            path = Environment.GetEnvironmentVariable(GlobalPackagesFolderEnvironmentKey);
+            // GlobalPackagesFolder path may be relative path. If so, it will be considered relative to
+            // the solution directory, just like the 'repositoryPath' setting
+            path = settings.GetValue(ConfigSection, GlobalPackagesFolderKey, isPath: false);
             if (!string.IsNullOrEmpty(path))
             {
                 path = path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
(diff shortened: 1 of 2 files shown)


src/NuGet.Configuration/Utility/SettingsUtility.cs
```diff
@@ -130,16 +130,16 @@ public static string GetGlobalPackagesFolder(ISettings settings)
                 throw new ArgumentNullException(nameof(settings));
             }
 
-            // Note that the value GlobalPackagesFolder must be a full path and not a relative path
-            // Read it like a string and explicitly not as a path
-            var path = settings.GetValue(ConfigSection, GlobalPackagesFolderKey, isPath: false);
+            var path = Environment.GetEnvironmentVariable(GlobalPackagesFolderEnvironmentKey);
             if (!string.IsNullOrEmpty(path))
             {
                 path = path.Replace(Path.AltDirectorySeparatorChar, Path.DirectorySeparatorChar);
```
emgarten, line 136: The replace/return should be done in one place, these can both use the same code path once it picks where to get the value from.

emgarten:
:shipit:

test/NuGet.Configuration.Test/SettingsTests.cs
```diff
@@ -1791,6 +1791,31 @@ public void GetGlobalPackagesFolder_FromNuGetConfig()
...
+</configuration>";
+
+            var nugetConfigPath = "NuGet.config";
+            using (var mockBaseDirectory = TestFilesystemUtility.CreateRandomTestFolder())
+            {
+                TestFilesystemUtility.CreateConfigurationFile(nugetConfigPath, mockBaseDirectory, config);
+                Settings settings = new Settings(mockBaseDirectory);
+
+                // Act
+                var globalPackagesFolderPath = SettingsUtility.GetGlobalPackagesFolder(settings);
```
pranavkm, line 1811: Does the calling code always `Path.GetFullPath(Path.Combine(Directory.GetCurrentDirectory(), SettingsUtility.GetGlobalPackagesFolder(settings)))`?

src/NuGet.Configuration/Utility/SettingsUtility.cs
pranavkm, line 136: Why do you need to do the replace at all? `Path` operations work well with either separators?

deepakaravindr, line 136: We need to replace it. We are reading in the value as a string, not as a path. Note that we have always done this for repositoryPath

test/NuGet.Configuration.Test/SettingsTests.cs
deepakaravindr, line 1811: No. With https://github.com/NuGet/Home/issues/1062, if globalPackagesFolder is relative, full path is constructed relative to solution directory. exe throws if solution directory is not available (Either the switch or.sln file)

test/NuGet.Configuration.Test/SettingsTests.cs
```diff
@@ -1791,6 +1791,31 @@ public void GetGlobalPackagesFolder_FromNuGetConfig()
...
+
+                // Assert
+                Assert.Equal(@"..\..\NuGetPackages", globalPackagesFolderPath);
```
yishaigalatzer, line 1814: is this test going to pass on Mono?

yishaigalatzer:
Can this PR be closed?

deepakaravindr:
Yes

PR #101, opened 2015-08-07
Don't wait for 401 - include authorization header if basic credential…

zarenner:
Resolves NuGet/Home#742 by always including Authorization header if basic credentials exist and the request is for the same server.

### DIFF
--- a/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
@@ -75,7 +75,7 @@ namespace NuGet.Protocol.Core.v3
                 CredentialStore.Instance.Add(uri, credential);
             }
 
-            return new CredentialPromptWebRequestHandler()
+            return new CredentialPromptWebRequestHandler(uri)
             {
                 Proxy = proxy,
                 Credentials = credential
@@ -84,12 +84,21 @@ namespace NuGet.Protocol.Core.v3
 
         private class CredentialPromptWebRequestHandler : WebRequestHandler
         {
+            private Uri _serviceIndexUri;
+
+            public CredentialPromptWebRequestHandler(Uri serviceIndexUri)
+            {
+                _serviceIndexUri = serviceIndexUri;
+            }
+
             protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
             {
                 while (true)
                 {
                     try
                     {
+                        AddBasicAuthorizationHeaderIfAvailable(request);
+
                         var response = await base.SendAsync(request, cancellationToken);
                         if (HttpHandlerResourceV3.ProxyPassed != null && Proxy != null)
                         {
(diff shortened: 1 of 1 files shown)


src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
```diff
@@ -121,6 +130,33 @@ private class CredentialPromptWebRequestHandler : WebRequestHandler
...
+            /// same scheme and authority as the service index.
+            /// </summary>
+            /// <remarks>
+            /// Resolves https://github.com/NuGet/Home/issues/742: NuGet should set PreAuthenticate to true to reduce the number of web requests
+            /// Note that PreAuthenticate doesn't do what you'd expect (it still doesn't use creds on the very first request),
+            /// so we set Authorization manually.
+            /// </remarks>
+            private void AddBasicAuthorizationHeaderIfAvailable(HttpRequestMessage request)
+            {
+                if (this.Credentials != null &&
```
yishaigalatzer, line 147: Please read and follow the coding standards for this repo

The short list - don't use this, use var, break long lines (if github shows the line to be longer than 120 chars break it), use using statements instead of specifically qualifying methods

src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
```diff
@@ -84,12 +84,21 @@ private HttpClientHandler TryGetCredentialAndProxy(PackageSource packageSource)
 
         private class CredentialPromptWebRequestHandler : WebRequestHandler
         {
+            private Uri _serviceIndexUri;
```
yishaigalatzer: readonly

src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
zarenner, line 147: Will do, thanks.

yishaigalatzer:
@zarenner thanks for this PR, We are going to hold on to it for the 3.2 release. We are not merging this for 3.1.2 release, there is not time to verify this or the urgency to take it. Please chime in if you disagree.

@blowdart can you please review this in terms of security? is this acceptable?

CC @feiling

blowdart:
It's fine by me

yishaigalatzer:
@zarenner please let me know once you have tests ready to go, and we can take it for 3.1.2 (which we are renaming to 3.2 today)

yishaigalatzer:
@zarenner is this PR still relevant after the recent changes?

zarenner:
Yes, it's still relevant in that with the existing code, we're still waiting for a 401 response before sending credentials.  This PR will (assuming the hostname of the request is the same as the package source) send the credentials without waiting for a 401.

See https://github.com/NuGet/Home/issues/742

There's no urgency or requirement to get this in now for us, but it's a perf improvement.

yishaigalatzer:
Based on our last discussion, this is probably incorrect to do (because of proxies).

Also the code is moving this week to NuGet.Client. Closing and lets have a separate discussion if we should resurrect this PR

PR #102, opened 2015-08-07
Lock file ordering

emgarten:
This change adds ordering to files, dependencies, and frameworks in the lock file to reduce the risk of merge conflicts between restores and different versions of NuGet when the lock file is added to source control.

Based on: https://github.com/aspnet/dnx/pull/2166

//cc @davidfowl @yishaigalatzer @deepakaravindr


deepakaravindr:
Looks good to me

PR #103, opened 2015-08-08
index.json cache refresh fix

emgarten:
This change fixes an issue where the service index was not being cached after the original cache entry expired.

https://github.com/NuGet/Home/issues/1054

//cc @deepakaravindr @pranavkm


deepakaravindr:
Looks good to me

PR #105, opened 2015-08-12
Setting the targetPath which is the fullPath to files added instead o…

deepakaravindr:
…f zipArchiveEntry's fullname. Related to https://github.com/NuGet/Home/issues/980

@yishaigalatzer @emgarten


emgarten:
:shipit:

PR #109, opened 2015-08-15
Retry on 5xx status codes

emgarten:
This change adds the RetryHandler back into the DataClient message handler. 5xx status codes and exceptions will be retried as part of the message handler.

Retries will be: 5 total tries with a 200ms wait in between each try (open to suggestions)

//cc @deepakaravindr @feiling @MeniZalzman @yishaigalatzer @johnataylor

### DIFF
--- a/src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
+++ b/src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
@@ -3,17 +3,14 @@
 
 using System;
 using System.Collections.Generic;
-using System.Globalization;
 using System.Linq;
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
 using NuGet.Protocol.Core.Types;
-using Newtonsoft.Json;
 #if !DNXCORE50
 using System.Net.Cache;
-
 #endif
 
 namespace NuGet.Protocol.Core.v3.Data
@@ -27,6 +24,14 @@ namespace NuGet.Protocol.Core.v3.Data
         private readonly INuGetMessageHandlerProvider[] _modifiers;
 
         /// <summary>
+        /// Use a HttpHandlerResource containing a message handler.
+        /// </summary>
+        public DataClient(HttpHandlerResource handlerResource)
+            : this(handlerResource.MessageHandler)
+        {
+        }
+
+        /// <summary>
         /// Raw constructor that allows full overriding of all caching and default DataClient behavior.
         /// </summary>
         public DataClient(HttpMessageHandler handler)
@@ -34,7 +39,7 @@ namespace NuGet.Protocol.Core.v3.Data
         {
             if (handler == null)
             {
-                throw new ArgumentNullException("handler");
+                throw new ArgumentNullException(nameof(handler));
             }
 
             // Set user agent
@@ -46,7 +51,7 @@ namespace NuGet.Protocol.Core.v3.Data
         /// DataClient with the default options and caching support
         /// </summary>
         public DataClient()
-            : this(CachingHandler)
+            : this(DefaultHandler)
         {
         }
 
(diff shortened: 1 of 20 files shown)


yishaigalatzer:
CC @davidfowl @lodejard

test/NuGet.Protocol.Core.v3.Tests/RetryHandlerTests.cs
```diff
@@ -0,0 +1,160 @@
...
+        public async Task RetryHandler_MessageHandlerMultipleTriesNoSuccess()
+        {
+            // Arrange
+            var hits = 0;
+
+            Func<HttpRequestMessage, HttpResponseMessage> handler = (request) =>
+            {
+                hits++;
+
+                return new HttpResponseMessage(System.Net.HttpStatusCode.ServiceUnavailable);
```
yishaigalatzer: add the using on top if possible

test/NuGet.Protocol.Core.v3.Tests/RetryHandlerTests.cs
```diff
@@ -0,0 +1,160 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using NuGet.Protocol.Core.Types;
+using NuGet.Protocol.Core.v3.Data;
+using Xunit;
+
+namespace NuGet.Protocol.Core.v3.Tests
+{
+    public class RetryHandlerTests
+    {
```
yishaigalatzer, line 14: Add a test to verify there is a minimal delay between the retries, move the delay calculation to a method, and verify the minimal delay by calling the method

Move the retry count to be a public readonly and use that when testing the max number of retries

yishaigalatzer:
Overall looks good, we also need more of a functional test to verify all restore scenarios actually restore from a mock server

Scenarios
V2 + package.config
V2 + project.json
V3 + packages.config
V3 + project.json

emgarten:
@yishaigalatzer there are powershell functional tests that verify restore for packages.config and project.json

yishaigalatzer:
Where are the tests that verify restore retries appropriately against a real server for all the above scenarios?

emgarten:
What those tests be needed verify that the retry code path is being hit for those scenarios? Or that the retry results are the same in an integration test?

yishaigalatzer:
That there are actual retries being hit, with appropriate delays in an integration test. we kept missing scenarios because of the multiple code paths in restore, an integration test will validate that. I'd rather that we run it against a mock server so its a. fast b. reproducible

emgarten:
Sounds good, I'll add mock server tests to cover this at the command line/PM layer.

emgarten:
I've added tests to verify the retry wait times.

src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
```diff
@@ -62,24 +70,33 @@ internal DataClient(HttpClientHandler handler, IEnumerable<INuGetMessageHandlerP
...
+            var messageHandler = new RetryHandler(clientHandler, 5);
 
-            //return modHandler;
+            return new HttpHandlerResourceV3(clientHandler, messageHandler);
+        }
 
+        /// <summary>
+        /// Chain the handlers together
+        /// </summary>
+        private static HttpClientHandler AssembleHandlers(HttpClientHandler handler, IEnumerable<INuGetMessageHandlerProvider> modifiers)
```
deepakaravindr: Looks like this private static method simply returns the HttpClientHandler 'handler'. Why does it take in an IEnumerable<INuGetMessageHandlerProvider>?

src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
```diff
@@ -27,6 +27,14 @@ public sealed class DataClient : HttpClient
         private readonly INuGetMessageHandlerProvider[] _modifiers;
```
deepakaravindr, line 24: Looks like this field is not used anywhere in the class today. If this is needed for the future, I would consider filing a bug to track this. If not, I will delete it

deepakaravindr, line 24: Based on the fact that HttpHandlerResourceV3 has a HttpMessageHandler for retry and a HttpClientHandler for credentials, IMO, I don't see the need for INuGetMessageHandlerProvider. Any need for extensibility should likely be accomplished using HttpMessageHandler(s) that can be configured

deepakaravindr:
:shipit: after incorporating the comments

emgarten:
I've removed modifiers from the assemble method. I am leaving the modifier classes in until we finalize the story for extensibility here. The original plan was to allow message handlers only, but there was push back on that since it was too complicated for simple scenarios.

src/NuGet.Protocol.Core.v3/DataClient/RetryHandler.cs
```diff
@@ -26,28 +31,37 @@ public RetryHandler(HttpMessageHandler innerHandler, int retries)
...
+            while (tries < MaxTries
                    && !success)
             {
-                // wait progressively longer
+                // wait before trying again if the last call failed
                 if (!success
                     && tries > 0)
                 {
-                    await Task.Delay(500 * tries, cancellationToken);
+                    await Task.Delay(200, cancellationToken);
```
MeniZalzman, line 41: Why you are not using the Property (RetryDelay) here?

emgarten, line 41: Good catch, RetryDelay was just added for the tests. I'll get this hooked into the actual code.

PR #110, opened 2015-08-17
Update the dependency on NuGet.Core to 2.8.8 so that the fix of NuGet/Home#776 gets incorporated into nuget v3

feiling:
Super simple change.

@zhili1208 @yishaigalatzer @emgarten @deepakaravindr


emgarten:
Make sure this builds across all repos and that PM and VS are updated also.
:shipit:

PR #114, opened 2015-08-21
fixed https://github.com/NuGet/Home/issues/1039

zhili1208:
https://github.com/NuGet/Home/issues/1039
won't fix the error message should be associated with project in error List here.
added project name to error message and handle the exception

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -149,7 +149,7 @@ namespace NuGet.Commands
         {
             if (_request.Project.TargetFrameworks.Count == 0)
             {
-                _log.LogError(Strings.Log_ProjectDoesNotSpecifyTargetFrameworks);
+                _log.LogError(Strings.FormatLog_ProjectDoesNotSpecifyTargetFrameworks(_request.Project.Name));
                 _success = false;
                 return Enumerable.Empty<RestoreTargetGraph>();
             }
@@ -400,7 +400,7 @@ namespace NuGet.Commands
         {
             // Get the project graph
             var projectFrameworks = project.TargetFrameworks.Select(f => f.FrameworkName).ToList();
-            if (projectFrameworks.Count > 1)
+            if (projectFrameworks.Count > 1 || targetGraphs.Count() == 0)
             {
                 return new MSBuildRestoreResult(project.Name, project.BaseDirectory);
             }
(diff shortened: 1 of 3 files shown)


zhili1208:
@emgarten @yishaigalatzer @feiling @deepakaravindr

src/NuGet.Commands/Strings.resx
```diff
@@ -151,7 +151,7 @@
     <value>Some packages are not compatible with {0}.</value>
   </data>
   <data name="Log_ProjectDoesNotSpecifyTargetFrameworks" xml:space="preserve">
-    <value>The project does not specify any target frameworks.</value>
+    <value>The project {0} does not specify any target frameworks in project.json.</value>
```
emgarten: Add single quotes around the project name like this '{0}'   It makes it easier to understand the message when the project has an odd name, or a name that blends into the sentence.

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -400,7 +400,7 @@ private bool ResolutionSucceeded(IEnumerable<RestoreTargetGraph> graphs)
         {
             // Get the project graph
             var projectFrameworks = project.TargetFrameworks.Select(f => f.FrameworkName).ToList();
-            if (projectFrameworks.Count > 1)
+            if (projectFrameworks.Count > 1 || targetGraphs.Count() == 0)
```
emgarten: You don't need to enumerate the whole list here, you can use: !targetGraphs.Any()

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -149,7 +149,7 @@ public Task<RestoreResult> ExecuteAsync()
         {
             if (_request.Project.TargetFrameworks.Count == 0)
             {
-                _log.LogError(Strings.Log_ProjectDoesNotSpecifyTargetFrameworks);
+                _log.LogError(Strings.FormatLog_ProjectDoesNotSpecifyTargetFrameworks(_request.Project.Name));
```
emgarten, line 152: lowercase s on string as per the convention. 

What is doing the formatting here? is it using the current culture?

yishaigalatzer:
String changes need pre-approval. I'm ok with this one

zhili1208:
@emgarten Yes, it's using the current culture, 
string.Format(CultureInfo.CurrentCulture, GetString("Log_ProjectDoesNotSpecifyTargetFrameworks"), p0);

emgarten:
:shipit:

PR #115, opened 2015-08-24
fixed https://github.com/NuGet/Home/issues/1172

zhili1208:
https://github.com/NuGet/Home/issues/1172
the root cause for this bug is package frameworks contain duplicate frameworks, that make FrameworkNameProvider can't match the profile table to get profile number. in this case, they are win and win8. 
this fix is to remove duplicate frameworks.

### DIFF
--- a/test/NuGet.Frameworks.Test/NuGetFrameworkParseTests.cs
+++ b/test/NuGet.Frameworks.Test/NuGetFrameworkParseTests.cs
@@ -255,10 +255,12 @@ namespace NuGet.Test
         [InlineData("portable-net45+wp8+win8+wpa")]
         [InlineData("portable-net45+win8+wp8+wpa81+monotouch+monoandroid")]
         [InlineData(".NETPortable,Version=v0.0,Profile=Profile259")]
-        // TODO: should bad framework names be supported?
-        //[InlineData(".NETPortable,Version=v0.0,Profile=net45+wp8+win8+wpa+monotouch+monoandroid")]
-        //[InlineData(".NETPortable,Version=v0.0,Profile=net45+wp8+win+wpa")]
-        //[InlineData(".NETPortable,Version=v0.0,Profile=net45+wp8+win8+wpa81")]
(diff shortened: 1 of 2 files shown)


test/NuGet.Frameworks.Test/NuGetFrameworkParseTests.cs
```diff
@@ -255,6 +255,7 @@ public void NuGetFramework_ParsePCLTest()
         [InlineData("portable-net45+wp8+win8+wpa")]
         [InlineData("portable-net45+win8+wp8+wpa81+monotouch+monoandroid")]
         [InlineData(".NETPortable,Version=v0.0,Profile=Profile259")]
+        [InlineData("portable-net45+wp8+win+wpa+win8")]
```
emgarten, line 258: This needs more tests. Vary and duplicate other frameworks. It should also check equivalent frameworks such as netcore45 instead of win8.

portable-net45+wp8+win+wpa+netcore+netcore45
portable-net450+net4.5+net45+wp8+wpa+win8+wpa81
portable-win8+net45+wp8+wpa81+win8+win8

emgarten:
This looks like it would work, but the fix duplicates a lot of code. It would be better to work this into GetEquivalentPermutations where it is already retrieving the equivalent frameworks. 

NuGetFramework.Parse() is hit a very high number of times so this fix needs to be as optimized as possible. 

:watch:

src/NuGet.Frameworks/DefaultFrameworkMappings.cs
```diff
@@ -194,6 +194,16 @@ public IEnumerable<FrameworkSpecificMapping> ProfileShortNames
                                 FrameworkConstants.CommonFrameworks.Dnx,
                                 FrameworkConstants.CommonFrameworks.Dnx451),
 
+                            // dnx451 <-> aspnet50
```
emgarten: why did these move? If the fix depends on the order here we need another fix.

emgarten:
:shipit:

PR #116, opened 2015-08-24
If the service index is available, it means that is a v3 package source,

deepakaravindr:
so always returns a ListCommandResource. If the list endpoint is not
available on the service index, ListCommandResource returned will have
a null endpoint

@yishaigalatzer @emgarten @MeniZalzman @feiling


emgarten:
:shipit:

MeniZalzman:
:shipit:

deepakaravindr:
Merged with https://github.com/NuGet/NuGet3/commit/35bd576fc8b3389475f831558581f86cee58ae14

PR #117, opened 2015-08-25
If the service index is available, it means that is a v3 package source,

deepakaravindr:
so always returns a PushCommandResource. If the push endpoint is not
available on the service index, PushCommandResource returned will have
a null endpoint

@yishaigalatzer @emgarten @MeniZalzman @feiling


emgarten:
:rocket:

MeniZalzman:
:shipit:

deepakaravindr:
Merged with https://github.com/NuGet/NuGet3/commit/dbba1caf34df3a96af25a186755198ea3493e5cb

PR #118, opened 2015-08-25
Fix NuGet/Home#1216: NuGet incorrectly treating authenticated V3 feed as V2

feiling:
The problem is that authentication was not supported for V3 feed. So when nuget checks
if a feed is V3 or not, since accessing index.json fails because of authentication failure,
nuget will treat the feed as V2.

This change added authentication support.

### DIFF
--- a/src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
@@ -3,11 +3,12 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Net.Http;
+using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
+using NuGet.Configuration;
 using NuGet.Protocol.Core.Types;
 using NuGet.Protocol.Core.v3.Data;
 using NuGet.Versioning;
@@ -37,10 +38,65 @@ namespace NuGet.Protocol.Core.v3
             MaxCacheDuration = _defaultCacheDuration;
         }
 
+        // Read the source's end point to get the index json.
+        // Returns null when there is a failure.
+        private async Task<JObject> GetIndexJson(SourceRepository source, CancellationToken token)
+        {
+            var uri = new Uri(source.PackageSource.Source);
+            ICredentials credentials = CredentialStore.Instance.GetCredentials(uri);
+            while (true)
+            {
+                var messageHandlerResource = await source.GetResourceAsync<HttpHandlerResource>(token);
+                if (credentials != null)
+                {
+                    messageHandlerResource.ClientHandler.Credentials = credentials;
+                }
+
+                using (var client = new DataClient(messageHandlerResource))
+                {
+                    var response = await client.GetAsync(uri, token);
+
+                    if (response.IsSuccessStatusCode)
+                    {
+                        if (HttpHandlerResourceV3.CredentialsSuccessfullyUsed != null && credentials != null)
+                        {
+                            HttpHandlerResourceV3.CredentialsSuccessfullyUsed(uri, credentials);
+                        }
+
+                        try
+                        {
+                            var text = await response.Content.ReadAsStringAsync();
+                            return JObject.Parse(text);
+                        }
+                        catch (JsonReaderException)
+                        {
+                            return null;
+                        }
+                    }
+                    else if (response.StatusCode == HttpStatusCode.Unauthorized)
+                    {
+                        credentials = null;
+                        if (HttpHandlerResourceV3.PromptForCredentials != null)
+                        {
+                            credentials = HttpHandlerResourceV3.PromptForCredentials(uri);
+                        }
+
+                        if (credentials == null)
+                        {
+                            return null;
+                        }
+                    }
+                    else
+                    {
+                        return null;
+                    }
+                }
+            }
+        }
+
         public override async Task<Tuple<bool, INuGetResource>> TryCreate(SourceRepository source, CancellationToken token)
         {
             ServiceIndexResourceV3 index = null;
-
             var url = source.PackageSource.Source;
 
             // the file type can easily rule out if we need to request the url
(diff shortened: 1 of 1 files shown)


src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
```diff
@@ -124,6 +124,11 @@ public JObject GetJObject(Uri address)
         public async Task<JObject> GetJObjectAsync(Uri address, CancellationToken token)
         {
             var response = await GetAsync(address, token);
+            if (response.StatusCode == System.Net.HttpStatusCode.Unauthorized)
+            {
+                throw new UnauthorizedAccessException();
+            }
+
```
feiling: The first change is that this method throws UnauthorizedAccessException() on authentication failure.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -56,29 +59,61 @@ public ServiceIndexResourceV3Provider()
...
+                            }
+
+                            if (credentials == null)
+                            {
+                                var entry = new ServiceIndexCacheInfo { CachedTime = utcNow };
+                                _cache.AddOrUpdate(url, entry, (key, value) => entry);
+
+                                return new Tuple<bool, INuGetResource>(false, null);
+                            }
+                        }
```
feiling, line 87: The change here is to catch UnauthorizedAccessException, prompt user for new credentials, then retry the request with the new credentials.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -44,6 +46,7 @@ public ServiceIndexResourceV3Provider()
             var url = source.PackageSource.Source;
 
             // the file type can easily rule out if we need to request the url
+
```
emgarten: Remove the new line

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -56,29 +59,61 @@ public ServiceIndexResourceV3Provider()
...
                     entryValidCutoff > cacheInfo.CachedTime)
                 {
-                    var messageHandlerResource = await source.GetResourceAsync<HttpHandlerResource>(token);
-
-                    var client = new DataClient(messageHandlerResource);
-
-                    JObject json;
+                    var uri = new Uri(url);
+                    ICredentials credentials = CredentialStore.Instance.GetCredentials(uri);
+                    JObject json = null;
```
emgarten: var

src/NuGet.Protocol.Core.v3/DataClient/DataClient.cs
```diff
@@ -124,6 +124,11 @@ public JObject GetJObject(Uri address)
         public async Task<JObject> GetJObjectAsync(Uri address, CancellationToken token)
         {
             var response = await GetAsync(address, token);
+            if (response.StatusCode == System.Net.HttpStatusCode.Unauthorized)
```
emgarten: Other resources are calling GetJObjectAsync and would not be expecting this. Either update the other resources to handle this, or use GetAsync instead and check the status code directly in the index.json resource.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -56,29 +59,61 @@ public ServiceIndexResourceV3Provider()
...
 
-                        return new Tuple<bool, INuGetResource>(false, null);
+                            return new Tuple<bool, INuGetResource>(false, null);
+                        }
+                        catch (HttpRequestException)
+                        {
+                            var entry = new ServiceIndexCacheInfo { CachedTime = utcNow };
+                            _cache.AddOrUpdate(url, entry, (key, value) => entry);
+
+                            return new Tuple<bool, INuGetResource>(false, null);
```
emgarten: This will lead to a large number of requests for index.json when a v3 source is unreachable or the user is offline. Please verify that this doesn't cause hangs and that we really want to keep retrying this.

emgarten: This will lead to a large number of requests for index.json when a v3 source is unreachable or the user is offline. Please verify that this doesn't cause hangs and that we really want to keep retrying this.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -56,29 +59,61 @@ public ServiceIndexResourceV3Provider()
...
+                    var uri = new Uri(url);
+                    ICredentials credentials = CredentialStore.Instance.GetCredentials(uri);
+                    JObject json = null;
 
-                    try
-                    {
-                        json = await client.GetJObjectAsync(new Uri(url), token);
-                    }
-                    catch (JsonReaderException)
+                    while (true)
```
emgarten: Retrying forever looks dangerous. If the wrong credentials are set in nuget.config the credentials will never be null it looks like this may go into an infinite loop.

emgarten:
:watch:

This won't actually stop the fallback to V2 if the user clicks cancel to the credential prompt or enters a blank user name. It will still return null and continue on to the v2 resource.

We should have the same auth logic on all resources instead of special casing index.json. I would expect the retries here to be on the dataclient or in the message handler. Checking the status code to determine if the json should be cached should be the only change here.

feiling:
Refactored the code to make it cleaner. Also addressed Justin's comments.
- _Other resources are calling GetJObjectAsync and would not be expecting this. Either update the other resources to handle this, or use GetAsync instead and check the status code directly in the index.json resource._ 

Very valid. Fixed. Reverted the change in GetJObjectAsync. The method now calls GetAsync() instead.
- _This will lead to a large number of requests for index.json when a v3 source is unreachable or the user is offline. Please verify that this doesn't cause hangs and that we really want to keep retrying this._

This is existing behavior. In fact, if v3 is unreachable, you can see that the code adds null to its cache so the network call won't be tried the next time this is called.
- _This won't actually stop the fallback to V2 if the user clicks cancel to the credential prompt or enters a blank user name. It will still return null and continue on to the v2 resource._

There is an open issue (NuGet/Home#1253) tracking the fall thru problem. 
- _Retrying forever looks dangerous. If the wrong credentials are set in nuget.config the credentials will never be null it looks like this may go into an infinite loop._

There is no infinite loop. If user clicks cancel, then `credentials` will be null. Note that there is an IF checking this condition and returns in this case.
- _We should have the same auth logic on all resources instead of special casing index.json. I would expect the retries here to be on the dataclient or in the message handler. Checking the status code to determine if the json should be cached should be the only change here._ 

Yes, it will be much better if authentication is handled in one central place. Unfortunately this is quite hard to do without making major changes. You can't do retry in DataClient. If you try it, you'll get an error saying there is an existing request. Instead you have to create a new DataClient for each new request.

emgarten:
:shipit:

PR #120, opened 2015-08-26
Cache control for project.json restore

emgarten:
This change adds a CacheContext to RestoreRequest to provide more options beyond NoCache. For package installs and updates restore will go by the time the package operation was started, all older cache entries for the packages restored will be cleared out of the v3-cache. After the operation rebuilding and restoring from the command line will work using the updated v3-cache content, instead of having to retrieving it again which would be the case with NoCache=true.

https://github.com/NuGet/Home/issues/1096

//cc @yishaigalatzer @davidfowl @MeniZalzman @deepakaravindr @feiling

### DIFF
--- a/src/NuGet.Protocol.Core.Types/SourceCacheContext.cs
+++ b/src/NuGet.Protocol.Core.Types/SourceCacheContext.cs
@@ -1,0 +1,95 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System;
+using System.Diagnostics;
+
+namespace NuGet.Protocol.Core.Types
+{
+    /// <summary>
+    /// Cache control settings for the V3 disk cache.
+    /// </summary>
+    public class SourceCacheContext
+    {
+        /// <summary>
+        /// Default amount of time to cache version lists.
+        /// </summary>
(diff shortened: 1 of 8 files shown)


MeniZalzman:
Can we add tests for this change?

MeniZalzman:
:shipit:

src/NuGet.Protocol.Core.Types/SourceCacheContext.cs
```diff
@@ -0,0 +1,95 @@
...
+
+        /// <summary>
+        /// Packages from the server older than this time span
+        /// will be fetched from the server.
+        /// </summary>
+        public TimeSpan NupkgMaxAgeTimeSpan
+        {
+            get
+            {
+                return GetCacheTime(ListMaxAge, DefaultCacheAgeLimitNupkg);
```
deepakaravindr, line 65: This should be NupkgMaxAge

src/NuGet.Protocol.Core.Types/SourceCacheContext.cs
```diff
@@ -0,0 +1,95 @@
...
+        {
+            var timeSpan = TimeSpan.Zero;
+
+            if (!NoCache)
+            {
+                // Default
+                timeSpan = defaultTime;
+
+                // If the max age is set use that instead of the default
+                if (ListMaxAge.HasValue)
```
deepakaravindr, line 79: This should be maxAge

deepakaravindr:
:shipit: after incorporating comments

PR #121, opened 2015-08-26
fixed https://github.com/NuGet/Home/issues/1163

zhili1208:
https://github.com/NuGet/Home/issues/1163
removed all those
Merging in runtimes defined in ...
Restoring packages for ...
Resolving conflicts for ...
Checking compatibility for ...

the new log will becomes 

capture

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -154,7 +154,7 @@ namespace NuGet.Commands
                 return Enumerable.Empty<RestoreTargetGraph>();
             }
 
-            _log.LogDebug(Strings.FormatLog_RestoringPackages(_request.Project.FilePath));
+            _log.LogInformation(Strings.FormatLog_RestoringPackages(_request.Project.FilePath));
 
             // Load repositories
             var projectResolver = new PackageSpecResolver(_request.Project);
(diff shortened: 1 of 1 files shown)


emgarten:
Lets fix this by filtering them at the console. It would be nice to keep these for debug purposes.
:watch:

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -154,7 +154,7 @@ public Task<RestoreResult> ExecuteAsync()
                 return Enumerable.Empty<RestoreTargetGraph>();
             }
 
-            _log.LogInformation(Strings.FormatLog_RestoringPackages(_request.Project.FilePath));
```
emgarten, line 157: Leaving this one

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -748,7 +748,7 @@ private void CheckDependencies(RestoreTargetGraph result, IEnumerable<LibraryDep
             var resultGraphs = new List<Task<RestoreTargetGraph>>();
             foreach (var runtimeName in runtimeIds)
             {
-                _log.LogInformation(Strings.FormatLog_RestoringPackages(FrameworkRuntimePair.GetName(graph.Framework, runtimeName)));
+                _log.LogDebug(Strings.FormatLog_RestoringPackages(FrameworkRuntimePair.GetName(graph.Framework, runtimeName)));
```
emgarten: Leave this also for the UI

emgarten:
Add back in the restore messages then :shipit:

deepakaravindr:
:shipit: after making the changes you suggested. Also, ensure that the behavior of 'nuget.exe restore <project.json>' for default VerbosityLevel is intact

yishaigalatzer:
@zhili1208 why is this PR still open?

PR #122, opened 2015-08-28
Fix of NuGet/Home#1253: Don't fall thru when v3 feed failed

feiling:
Now if there is a failure, an exception will get thrown so GetResource fails immediately so no fall thru will happens. Also the V3 resource is cached only when it is successfully created.

### DIFF
--- a/src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Concurrent;
+using System.Globalization;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -39,7 +40,7 @@ namespace NuGet.Protocol.Core.v3
         }
 
         // Read the source's end point to get the index json.
-        // Returns null when there is a failure.
+        // An exception will be thrown on failure.
         private async Task<JObject> GetIndexJson(SourceRepository source, CancellationToken token)
         {
             var uri = new Uri(source.PackageSource.Source);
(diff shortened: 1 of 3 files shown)


src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -83,12 +77,12 @@ public ServiceIndexResourceV3Provider()
...
                         if (credentials == null)
                         {
-                            return null;
+                            response.EnsureSuccessStatusCode();
                         }
                     }
                     else
                     {
-                        return null;
+                        response.EnsureSuccessStatusCode();
```
feiling, line 85: This, plus the previous EnsureSuccessStatusCode(), makes sure that this method returns only when Index.Json is successfully fetched and parsed, otherwise an exception will get thrown.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -113,26 +107,31 @@ public ServiceIndexResourceV3Provider()
...
+                            throw new NuGetProtocolException(errorMessage);
+                        }
                     }
                     else
                     {
-                        var entry = new ServiceIndexCacheInfo { CachedTime = utcNow };
-                        _cache.AddOrUpdate(url, entry, (key, value) => entry);
-
-                        return new Tuple<bool, INuGetResource>(false, null);
+                        throw new NuGetProtocolException(Strings.Protocol_MissingVersion);
```
feiling, line 134: This throw, plus the previous throw, makes sure that there is no exception only when index.json is legit.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -113,26 +107,31 @@ public ServiceIndexResourceV3Provider()
...
                     {
-                        // Use SemVer instead of NuGetVersion, the service index should always be
-                        // in strict SemVer format
                         SemanticVersion version;
-                        JToken versionToken;
-                        if (json.TryGetValue("version", out versionToken) &&
-                            versionToken.Type == JTokenType.String &&
-                            SemanticVersion.TryParse((string)versionToken, out version) &&
+                        if (SemanticVersion.TryParse((string)versionToken, out version) &&
                             version.Major == 3)
```
feiling, line 119: The checks were split so the user gets better error messages.

emgarten:
:shipit:

feiling:
Also fixed affected tests. Merged and checked in.

PR #123, opened 2015-08-28
PackagesConfigWriter update support

emgarten:
This change adds PackagesConfigWriter support for adding a new entry based on an entry from the original packages.config.

Previously unknown attributes were being dropped, this change also fixes that and persists all attributes except for require re-installation.

https://github.com/NuGet/Home/issues/1130

//cc @deepakaravindr @zhili1208 @MeniZalzman @yishaigalatzer

### DIFF
--- a/src/NuGet.Packaging/PackagesConfigWriter.cs
+++ b/src/NuGet.Packaging/PackagesConfigWriter.cs
@@ -388,14 +388,14 @@ namespace NuGet.Packaging
                 // When the attribute is not specified a value in the new node
                 if (string.IsNullOrEmpty(newValue))
                 {
-                    if (!name.Equals(XName.Get(PackagesConfig.allowedVersionsAttributeName)))
+                    if (name.Equals(XName.Get(PackagesConfig.RequireInstallAttributeName)))
                     {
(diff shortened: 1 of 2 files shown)


src/NuGet.Packaging/PackagesConfigWriter.cs
```diff
@@ -182,7 +182,49 @@ public void UpdatePackageEntry(PackageReference oldEntry, PackageReference newEn
                 matchingNode.ReplaceWith(newPackageNode);
             }
         }
-        
+
+        /// <summary>
+        /// Update a package entry using the original entry as a base if it exists.
+        /// </summary>
+        public void UpdatePackageEntry(XDocument originalConfig, PackageReference newEntry)
```
MeniZalzman: The Method should be named UpdateOrAddPackageEntry because if the new entry not exist you are adding her to the file anyway

MeniZalzman:
:watch:

src/NuGet.Packaging/PackagesConfigWriter.cs
emgarten: I'll update this

deepakaravindr:
:shipit:

PR #125, opened 2015-08-29
Fix library file slashes

emgarten:
The content model expects all slashes to be forward slashes. In some scenarios the existing lock file is used to skip reading packages from disk, and in these cases back slashes can get in and the content model will fail to recognize build items.

//cc @deepakaravindr @yishaigalatzer 

https://github.com/NuGet/Home/issues/1288


deepakaravindr:
Looks good to me. :shipit:

PR #126, opened 2015-08-31
Fix NuGet/Home#1300: V3 install/restore only uses credentials for index.json.

feiling:
The problem is that in HttpHandlerResourceV3Provider, when there is a credential saved in
nuget.config, the code creates CredentialCache, then adds NetworkCredential into the cache, with the
uri of the source as key. However, for V3 source, other resources most likely have different URIs
from index.json endpoint. Thus, this CredentialCache won't be able to provide credentials for other
resources. It only works for index.json.

The fix is to just create a normal NetworkCredential object.

### DIFF
--- a/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
@@ -14,8 +14,6 @@ namespace NuGet.Protocol.Core.v3
 {
     public class HttpHandlerResourceV3Provider : ResourceProvider
     {
-        private static readonly string[] _authenticationSchemes = new[] { "Basic", "NTLM", "Negotiate" };
-
         public HttpHandlerResourceV3Provider()
             : base(typeof(HttpHandlerResource),
                   nameof(HttpHandlerResourceV3Provider),
@@ -66,12 +64,7 @@ namespace NuGet.Protocol.Core.v3
                 && !String.IsNullOrEmpty(packageSource.UserName)
                 && !String.IsNullOrEmpty(packageSource.Password))
             {
-                var cache = new CredentialCache();
-                foreach (var scheme in _authenticationSchemes)
-                {
-                    cache.Add(uri, scheme, new NetworkCredential(packageSource.UserName, packageSource.Password));
-                }
-                credential = cache;
+                credential = new NetworkCredential(packageSource.UserName, packageSource.Password);
             }
 
             if (proxy != null)


src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
```diff
@@ -66,12 +64,7 @@ private HttpClientHandler TryGetCredentialAndProxy(PackageSource packageSource)
                 && !String.IsNullOrEmpty(packageSource.UserName)
                 && !String.IsNullOrEmpty(packageSource.Password))
             {
-                var cache = new CredentialCache();
```
yishaigalatzer, line 69: So with this we make host == account? Which means you can't host two private feeds on the same host? I think that's an ok assumption for now, we just have to document it (start a PR in the docs repo as well)

CC @csharpfritz

yishaigalatzer:
:shipit: when comments addressed and another signoff

src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
yishaigalatzer, line 69: Talking to the VSO folks, this will break some of their scenarios where the hosts are different between where index.json is hosted and where the registration blobs are at.

The suggestion is to make the code flow the credentials around. @zarenner is going to propose changes, and I'm going to ask @blowdart to take a look to see if we are worried about credential leak or not

@daazose

zarenner, line 69: Looking at this again, David and I may have misunderstood something in our discussion with Yishai.  My understanding now is that CredentialPromptWebRequestHandler gets non-domain-specific credentials, which it will happily use on all requests, regardless of the request hostname.  That same CredentialPromptWebRequestHandler is re-used in multiple requests which may be for different hostnames.

If we're still misunderstanding something, please correct me, but otherwise this change looks fine to us.  We're validating it now against our scenarios.

emgarten:
Is the username updated on the package source after prompting? Is this relying on the call to index.json to handle all retries for getting the right credentials?

yishaigalatzer:
What's the scenario where the password will differ?

Sent from my Windows Phone

src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
feiling, line 69: @yishaigalatzer Right, with the current implementation, you can't host two different feeds that require different credentials on the same host. We need to redesign our authentication implementation to support this scenario.

feiling:
@emgarten after prompting, the new username/password will get saved in CredentialStore.Instance. So the next time when this method is called, after this line

```
var credential = CredentialStore.Instance.GetCredentials(uri);
```

`credential` will be the new username/password, so this condition will be false:

```
if (credential == null
    && !String.IsNullOrEmpty(packageSource.UserName)
    && !String.IsNullOrEmpty(packageSource.Password))
```

and the username/password saved in nuget.config will NOT get used.

emgarten:
@yishaigalatzer when the nuget.config credentials are incorrect

@feiling :shipit:

yishaigalatzer:
I get it. My question was when can they be incorrect for anything index.json represents but correct for index.json

Sent from my Windows Phone

PR #127, opened 2015-09-03
Allow prerelease versions during resolver conflict checks

emgarten:
This change updates the resolver dependency conflict check to allow prerelease versions if one was selected during the graph walk. By default ranges with stable versions do not allow prerelease versions, but at this point in the code if a prerelease version was selected it means that the dependency constraint allowed it and it should be accepted here.

I've also added VersionRange.SetIncludePrerelease and cleaned up other parts of the code performing this same action.

Additional tests for this are in packagemanagement in the command line tests.

https://github.com/NuGet/Home/issues/1304

//cc @davidfowl @yishaigalatzer @deepakaravindr @MeniZalzman @feiling

### DIFF
--- a/src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
+++ b/src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
+using NuGet.Versioning;
 
 namespace NuGet.DependencyResolver
 {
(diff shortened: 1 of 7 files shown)


src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
```diff
@@ -260,14 +261,32 @@ private static bool TryResolveConflicts<TItem>(this GraphNode<TItem> root, List<
...
-                        versionConflicts.Add(new VersionConflictResult<TItem>
+                        var versionRange = childNode.Key.VersionRange;
+                        var checkVersion = acceptedNode.Item.Key.Version;
+
+                        // Allow prerelease versions if the selected library is prerelease and the range is
+                        // using the default behavior of filtering to stable versions.
+                        // Ex: [4.0.0, ) should allow 4.0.10-beta if that library was selected during the graph walk
+                        // The decision on if a prerelease version should be allowed should happen previous to this
+                        // check during the walk.
+                        if (checkVersion.IsPrerelease && !versionRange.IncludePrerelease)
```
deepakaravindr, line 277: Since VersionRange.SetIncludePrerelease already checks for versionRange.IncludePrerelease, we may not have to check it here

deepakaravindr:
:shipit:

src/NuGet.DependencyResolver.Core/GraphModel/GraphOperations.cs
emgarten, line 277: @deepakaravindr this code could just always set it to include prerelease and it would work since the filtering is done before hand. But what this if statement is really doing here is checking if the resolved package is prerelease before updating the range to allow prerelease. In SetIncludePrelease it does the check just to noop if it can, which is different.

PR #128, opened 2015-09-08
Add configurable package download timeout

zarenner:
Large package downloads currently time out at 1:40min due to the Timeout property on DataClient (HttpClient), which makes it impossible to restore very large packages.

This change adds a configurable property 'downloadTimeout' on a package source.

This change also increases the default to 5 minutes (instead of the 1:40 default in HttpClient).  Please let me know if you disagree with this default value.

### DIFF
--- a/src/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -12,6 +12,11 @@ namespace NuGet.Configuration
         /// </summary>
         public const int DefaultProtocolVersion = 2;
 
+        /// <summary>
+        /// The default timeout when downloading packages.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -44,6 +49,11 @@ public class PackageSource : IEquatable<PackageSource>
         /// </summary>
         public int ProtocolVersion { get; set; } = DefaultProtocolVersion;
 
+        /// <summary>
+        /// Gets or sets the maximum timeout in seconds when downloading packages from the source.
+        /// </summary>
+        public int DownloadTimeout { get; set; } = DefaultDownloadTimeout;
```
jmyersmsft: Please store time spans as TimeSpans

PR #129, opened 2015-09-10
Improve nuspec handling

emgarten:
This change improves nuspec handling and makes it consistent across packages.config and project.json projects.

The nuspec will be found by checking for .nuspec files only in the root of the zip. If only one exists it will be used. If zero or multiple exist a packaging exception will be thrown. Nuspecs in sub folders will be ignored.

PackageReader now finds the nuspec directly from the zip entries instead of using the GetFiles and GetStream abstractions which caused it to loop through all entries multiple times.

OpenNuspecFromNupkgAsync will now return a NuspecReader which reads the XML directly instead of copying the entry into a memory stream before reading it as XML.

https://github.com/NuGet/Home/issues/1358

//cc @deepakaravindr @feiling @MeniZalzman @yishaigalatzer

### DIFF
--- a/test/NuGet.Packaging.Test/PackageFolderReaderTests.cs
+++ b/test/NuGet.Packaging.Test/PackageFolderReaderTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
 
 using System;
+using System.Collections.Concurrent;
 using System.IO;
 using System.IO.Compression;
 using System.Linq;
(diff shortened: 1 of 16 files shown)


test/NuGet.Packaging.Test/PackageFolderReaderTests.cs
```diff
@@ -11,13 +12,259 @@
...
-    public class PackageFolderReaderTests
+    public class PackageFolderReaderTests : IDisposable
     {
+        [Fact]
+        public void PackageFolderReader_NuspecCountOne()
+        {
+            // Arrange
+            var workingDir = GetTempDir();
+
+            var stream = new MemoryStream();
```
deepakaravindr, line 23: Don't you want a using statement around the stream to dispose it. Same comment for all the new tests

deepakaravindr:
Looks good to me

yishaigalatzer:
CC @davidfowl

emgarten:
This is going in post-3.2.0

yishaigalatzer:
Need to move to NuGet.Client

emgarten:
It's already in NuGet.Client, I will close this since it will never go into NuGet3.

PR #104, opened 2015-08-11
Add Xamarin.WatchOS framework

mrward:
This adds a new Xamarin.WatchOS target framework to NuGet 3.

There is also a [pull request](https://nuget.codeplex.com/SourceControl/network/forks/MattWard/mononuget/contribution/8405) on CodePlex to add the new target framework to NuGet 2.8.8.

PR #130, opened 2015-09-11
Add Xamarin.tvOS framework

mrward:
This adds a new Xamarin.tvOS target framework to NuGet v3.

Also includes the Xamarin.WatchOS target framework which is part of an earlier [pull request](https://github.com/NuGet/NuGet3/pull/104).


zhili1208:
ported to NuGet.Client and merged it

PR #131, opened 2015-09-15
Set Default RequestCachePolicy

darrelmiller:
This handler is used when available in place of
DataClient.DefaultHandler which has caching set to Default.  It makes
sense to me that this handler should have the same settings.

### DIFF
--- a/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
+++ b/src/NuGet.Protocol.Core.v3/HttpHandlerResourceV3Provider.cs
@@ -85,6 +85,11 @@ namespace NuGet.Protocol.Core.v3
 
         private class CredentialPromptWebRequestHandler : WebRequestHandler
         {
+            public CredentialPromptWebRequestHandler()
+            {
+                CachePolicy = new System.Net.Cache.RequestCachePolicy(System.Net.Cache.RequestCacheLevel.Default);
+            }
+
             protected override async Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
             {
                 while (true)


yishaigalatzer:
@darrelmiller this looks like a good change, but we are going to hold on it for a bit before taking it. We want to make improvements to the network behavior without relying on caching. e.g. we now call index.json multiple times when installing a package, and even though caching will "fix" it, there is really no need to do it as part of a single download "transaction". In this case caching will make things appear to work fine, although we are doing to much work.

The other thing going on is repo restructuing. We plan to merge the repos building NuGet3, to simplify the build, debugging and testing process (which I'm sure you can appreciate :) ). We are half way through it, and I expect it to be done by end of next week or earlier, at that point you might have to rebase your code or move it alltogether.

Thanks again for another valueable contribution!

darrelmiller:
Ok.  Makes sense.  In the meanwhile if you are curious to see how awesomely quick the nuget dialog opens when caching is enabled, just add the following rule to Fiddler's OnPeekAtResponseHeaders event:

```
  if (oSession.uriContains("//api-v3search-0.nuget.org/query?q=&")) {
        oSession.ResponseHeaders.Remove("Cache-Control");
        oSession.ResponseHeaders.Remove("Pragma");
        oSession.ResponseHeaders.Remove("Expires");
        oSession.ResponseHeaders.Add("Cache-Control","private, max-age=60");
    }

    if (oSession.uriContains("//api.nuget.org/v3/index.json")) {
        oSession.ResponseHeaders.Remove("Cache-Control");
        oSession.ResponseHeaders.Add("Cache-Control","private, max-age=60");
    }

    if (oSession.uriContains("//api.nuget.org/v3/registration1/entityframework/index.json")) {
        oSession.ResponseHeaders.Remove("Cache-Control");
        oSession.ResponseHeaders.Add("Cache-Control","private, max-age=20");
    }
```

and add the following to machine.config to work around the Visual Studio issue,

```
 <system.net>
     <requestCaching defaultPolicyLevel="Default"/>
  </system.net>
```

emgarten:
Merged as NuGet/NuGet.Client@1974574a8b3b4df9b1690c2718ba7010b5f8f64e

PR #132, opened 2015-09-24
Fall back to runtime-agnostic native folder if package doesn't have a runtime-specific native folder

dsplaisted:
Fixes NuGet/Home#1454

### DIFF
--- a/src/NuGet.Client/ManagedCodeConventions.cs
+++ b/src/NuGet.Client/ManagedCodeConventions.cs
@@ -211,8 +211,16 @@ namespace NuGet.Client
             public SelectionCriteria ForRuntime(string runtimeIdentifier)
             {
                 var builder = new SelectionCriteriaBuilder(_conventions.Properties);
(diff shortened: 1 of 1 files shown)


yishaigalatzer:
Hey

1st please let us triage the bug and make a decision if we want this suggested fix or not :)

Second we are moving out of NuGet3 into nuget.client, we are not taking Pull Requests at this point in this repo because dev is open for 3.2.1 fixes only. So its a good time to take a look at it and make a decision, but the PR itself will need to go somewhere else.

yishaigalatzer:
@dsplaisted Based on our offline discussion, I believe this is not the right fix. Are we in agreement?

dsplaisted:
@yishaigalatzer I will set up a meeting (probably next week) for us to discuss what we're trying to do and what the right way to do it is.

yishaigalatzer:
@dsplaisted I'm closing this here, since we are no longer going to take PRs in this repo. All new work should go to nuget.client repo

PR #133, opened 2015-09-28
nuget.exe 3.2.1 perf improvements (part 1)

emgarten:
Perf improvements for the v2 and v3 protocol. V3 constructs urls, V2 makes use of IHttpClient directly for fine grained control over the download.
- Fixed a major threading issue with V2 online downloads. IPackage instances depend on the context of the repository they came from and this is not thread safe in NuGet.Core. When multiple packages were downloaded from V2 at the same time a race case could occur. The fix for this is to create a new repo for each operation. For simple downloads this does not cause extra calls.  https://github.com/NuGet/Home/issues/1448
- V2 downloads can now be canceled. Previously a download would run to completion on an abandoned thread and tie up resources.
- Added download url and package hash to SourcePackageDependencyInfo to avoid going back for the data again when the package is downloaded/installed. https://github.com/NuGet/Home/issues/1426
- DownloadResourceV2 now creates an IHttpClient directly and downloads the package.
- V2 downloads will now set the correct user agent.
- Download resources will use the stored download url for a package if it exists instead of looking the package up again against v2, or finding it a second time in the registration blobs.
- V3 downloads will construct the url for the flat container download if the index.json file contains a flat container resource. This allows direct downloads and avoids an extra lookup in the registration blob.

### DIFF
--- a/src/NuGet.Commands/RestoreCommand.cs
+++ b/src/NuGet.Commands/RestoreCommand.cs
@@ -8,7 +8,6 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
-using NuGet.Configuration;
 using NuGet.ContentModel;
 using NuGet.DependencyResolver;
 using NuGet.Frameworks;
@@ -868,12 +867,13 @@ namespace NuGet.Commands
                 token: token);
         }
 
-        private IRemoteDependencyProvider CreateProviderFromSource(PackageSource source, SourceCacheContext cacheContext)
+        private IRemoteDependencyProvider CreateProviderFromSource(
+            SourceRepository repository,
+            SourceCacheContext cacheContext)
         {
-            _log.LogVerbose(Strings.FormatLog_UsingSource(source.Source));
+            _log.LogVerbose(Strings.FormatLog_UsingSource(repository.PackageSource.Source));
 
-            var nugetRepository = Repository.Factory.GetCoreV3(source.Source);
-            return new SourceRepositoryDependencyProvider(nugetRepository, _log, cacheContext);
+            return new SourceRepositoryDependencyProvider(repository, _log, cacheContext);
         }
     }
 }
--- a/src/NuGet.Commands/RestoreRequest.cs
+++ b/src/NuGet.Commands/RestoreRequest.cs
@@ -8,6 +8,7 @@ using NuGet.Configuration;
 using NuGet.Frameworks;
 using NuGet.ProjectModel;
 using NuGet.Protocol.Core.Types;
+using NuGet.Protocol.Core.v3;
 
 namespace NuGet.Commands
 {
@@ -20,9 +21,35 @@ namespace NuGet.Commands
         { }
 
         public RestoreRequest(PackageSpec project, IEnumerable<PackageSource> sources, string packagesDirectory)
+            : this(project, packagesDirectory)
         {
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            Sources = sources.Select(source => Repository.Factory.GetCoreV3(source.Source)).ToList();
+        }
+
+        public RestoreRequest(PackageSpec project, IEnumerable<SourceRepository> sources, string packagesDirectory)
+            : this(project, packagesDirectory)
+        {
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            Sources = sources.ToList();
+        }
+
+        private RestoreRequest(PackageSpec project, string packagesDirectory)
+        {
+            if (project == null)
+            {
+                throw new ArgumentNullException(nameof(project));
+            }
+
             Project = project;
-            Sources = sources.ToList().AsReadOnly();
 
             ExternalProjects = new List<ExternalProjectReference>();
             CompatibilityProfiles = new HashSet<FrameworkRuntimePair>();
@@ -40,7 +67,7 @@ namespace NuGet.Commands
         /// <summary>
         /// The complete list of sources to retrieve packages from (excluding caches)
         /// </summary>
-        public IReadOnlyList<PackageSource> Sources { get; }
+        public IReadOnlyList<SourceRepository> Sources { get; }
 
         /// <summary>
         /// The directory in which to install packages
--- a/src/NuGet.Configuration/PackageSource/PackageSource.cs
+++ b/src/NuGet.Configuration/PackageSource/PackageSource.cs
@@ -14,6 +14,7 @@ namespace NuGet.Configuration
 
         private readonly int _hashCode;
         private bool? _isHttp;
+        private bool? _isLocal;
 
         public string Name { get; private set; }
 
@@ -55,6 +56,30 @@ namespace NuGet.Configuration
                 }
 
                 return _isHttp.Value;
+            }
+        }
+
+        /// <summary>
+        /// True if the source path is file based. Unc shares are not included.
+        /// </summary>
+        public bool IsLocal
+        {
+            get
+            {
+                if (!_isLocal.HasValue)
+                {
+                    Uri uri;
+                    if (Uri.TryCreate(Source, UriKind.Absolute, out uri))
+                    {
+                        _isLocal = uri.IsFile;
+                    }
+                    else
+                    {
+                        _isLocal = false;
+                    }
+                }
+
+                return _isLocal.Value;
             }
         }
 
--- a/src/NuGet.Protocol.Core.Types/SourcePackageDependencyInfo.cs
+++ b/src/NuGet.Protocol.Core.Types/SourcePackageDependencyInfo.cs
@@ -1,6 +1,7 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
 using NuGet.Packaging.Core;
-using NuGet.Protocol.Core.Types;
 using NuGet.Versioning;
 
 namespace NuGet.Protocol.Core.Types
@@ -8,28 +9,100 @@ namespace NuGet.Protocol.Core.Types
     public class SourcePackageDependencyInfo : PackageDependencyInfo
     {
         public SourcePackageDependencyInfo(string id, NuGetVersion version, bool listed, SourceRepository source)
-            : base(id, version)
+            : this(
+                  id,
+                  version,
+                  dependencies: Enumerable.Empty<PackageDependency>(),
+                  listed: listed,
+                  source: source)
         {
-            Listed = listed;
-            Source = source;
         }
 
-        public SourcePackageDependencyInfo(PackageIdentity identity, IEnumerable<PackageDependency> dependencies, bool listed, SourceRepository source)
+        public SourcePackageDependencyInfo(
+            string id,
+            NuGetVersion version,
+            IEnumerable<PackageDependency> dependencies,
+            bool listed,
+            SourceRepository source)
+            : this(
+                  new PackageIdentity(id, version),
+                  dependencies,
+                  listed,
+                  source,
+                  downloadUri: null,
+                  packageHash: null)
+        {
+        }
+
+        public SourcePackageDependencyInfo(
+            string id,
+            NuGetVersion version,
+            IEnumerable<PackageDependency> dependencies,
+            bool listed,
+            SourceRepository source,
+            Uri downloadUri,
+            string packageHash)
+            : this(
+                  new PackageIdentity(id, version),
+                  dependencies,
+                  listed,
+                  source,
+                  downloadUri,
+                  packageHash)
+        {
+        }
+
+        public SourcePackageDependencyInfo(
+            PackageIdentity identity,
+            IEnumerable<PackageDependency> dependencies,
+            bool listed,
+            SourceRepository source)
+            : this(
+                  identity,
+                  dependencies,
+                  listed,
+                  source,
+                  downloadUri: null,
+                  packageHash: null)
+        {
+        }
+
+        public SourcePackageDependencyInfo(
+            PackageIdentity identity,
+            IEnumerable<PackageDependency> dependencies,
+            bool listed,
+            SourceRepository source,
+            Uri downloadUri,
+            string packageHash)
             : base(identity, dependencies)
         {
             Listed = listed;
             Source = source;
+            DownloadUri = downloadUri;
+            PackageHash = packageHash;
         }
 
-        public SourcePackageDependencyInfo(string id, NuGetVersion version, IEnumerable<PackageDependency> dependencies, bool listed, SourceRepository source)
-            : base(id, version, dependencies)
-        {
-            Listed = listed;
-            Source = source;
-        }
-
+        /// <summary>
+        /// True if the package is listed and shown in search.
+        /// </summary>
+        /// <remarks>This property only applies to online sources.</remarks>
         public bool Listed { get; }
 
+        /// <summary>
+        /// Source repository the dependency information was retrieved from.
+        /// </summary>
         public SourceRepository Source { get; }
+
+        /// <summary>
+        /// The HTTP, UNC, or local file URI to the package nupkg.
+        /// </summary>
+        /// <remarks>Optional</remarks>
+        public Uri DownloadUri { get; }
+
+        /// <summary>
+        /// Package hash
+        /// </summary>
+        /// <remarks>Optional</remarks>
+        public string PackageHash { get; }
     }
 }
--- a/src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
+++ b/src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
@@ -63,7 +63,7 @@ namespace NuGet.Protocol.Core.v2
                 try
                 {
                     // Retrieve all packages
-                    var repoPackage = V2Client.FindPackage(package.Id, legacyVersion, allowPrereleaseVersions: true, allowUnlisted: true);
+                    var repoPackage = GetSource().FindPackage(package.Id, legacyVersion, allowPrereleaseVersions: true, allowUnlisted: true);
 
                     if (repoPackage != null)
                     {
@@ -110,7 +110,7 @@ namespace NuGet.Protocol.Core.v2
             try
             {
                 // Retrieve all packages
-                var repoPackages = V2Client.FindPackagesById(packageId);
+                var repoPackages = GetSource().FindPackagesById(packageId);
 
                 // Convert from v2 to v3 types and enumerate the list to finish all server requests before returning
                 results = repoPackages.Select(p => CreateDependencyInfo(p, projectFramework)).ToList();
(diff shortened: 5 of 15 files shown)


emgarten:
//cc @deepakaravindr @feiling @johnataylor @yishaigalatzer @MeniZalzman

src/NuGet.Commands/RestoreCommand.cs
```diff
@@ -868,12 +867,13 @@ private RuntimeGraph LoadRuntimeGraphCore(LocalPackageInfo package)
                 token: token);
         }
 
-        private IRemoteDependencyProvider CreateProviderFromSource(PackageSource source, SourceCacheContext cacheContext)
+        private IRemoteDependencyProvider CreateProviderFromSource(
+            SourceRepository repository,
+            SourceCacheContext cacheContext)
         {
-            _log.LogVerbose(Strings.FormatLog_UsingSource(source.Source));
+            _log.LogVerbose(Strings.FormatLog_UsingSource(repository.PackageSource.Source));
```
yishaigalatzer, line 874: Do we have a guarantee that repository.PackageSource is not null?

src/NuGet.Commands/RestoreRequest.cs
```diff
@@ -20,9 +21,35 @@ public RestoreRequest(PackageSpec project, IEnumerable<PackageSource> sources)
         { }
 
         public RestoreRequest(PackageSpec project, IEnumerable<PackageSource> sources, string packagesDirectory)
+            : this(project, packagesDirectory)
```
yishaigalatzer, line 24: XMLDocs? (And on top)

src/NuGet.Commands/RestoreRequest.cs
```diff
@@ -20,9 +21,35 @@ public RestoreRequest(PackageSpec project, IEnumerable<PackageSource> sources)
...
         {
+            if (sources == null)
+            {
+                throw new ArgumentNullException(nameof(sources));
+            }
+
+            Sources = sources.Select(source => Repository.Factory.GetCoreV3(source.Source)).ToList();
```
yishaigalatzer, line 31: Not sure I follow, what does GetCoreV3 do? Is it just that the name is misleading?

src/NuGet.Commands/RestoreRequest.cs
```diff
@@ -20,9 +21,35 @@ public RestoreRequest(PackageSpec project, IEnumerable<PackageSource> sources)
...
+        }
+
+        public RestoreRequest(PackageSpec project, IEnumerable<SourceRepository> sources, string packagesDirectory)
```
yishaigalatzer, line 34: There are too many constructors, with too many ToList calls. Is it possible to simplify it

src/NuGet.Configuration/PackageSource/PackageSource.cs
```diff
@@ -58,6 +59,30 @@ public bool IsHttp
...
         }
 
+        /// <summary>
+        /// True if the source path is file based. Unc shares are not included.
+        /// </summary>
+        public bool IsLocal
+        {
+            get
+            {
+                if (!_isLocal.HasValue)
```
yishaigalatzer, line 69: Why lazy compute rather than explicitly compute on creation?

src/NuGet.Protocol.Core.Types/SourcePackageDependencyInfo.cs
```diff
@@ -8,28 +9,100 @@ namespace NuGet.Protocol.Core.Types
     public class SourcePackageDependencyInfo : PackageDependencyInfo
     {
         public SourcePackageDependencyInfo(string id, NuGetVersion version, bool listed, SourceRepository source)
-            : base(id, version)
+            : this(
+                  id,
```
yishaigalatzer, line 13: nit: id should go on the same line as `this(`

src/NuGet.Protocol.Core.Types/SourcePackageDependencyInfo.cs
```diff
@@ -8,28 +9,100 @@ namespace NuGet.Protocol.Core.Types
...
+                  listed: listed,
+                  source: source)
         {
-            Listed = listed;
-            Source = source;
         }
 
-        public SourcePackageDependencyInfo(PackageIdentity identity, IEnumerable<PackageDependency> dependencies, bool listed, SourceRepository source)
-            : base(identity, dependencies)
+        public SourcePackageDependencyInfo(
```
yishaigalatzer, line 21: Do we truly need that many constructors?

src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
```diff
@@ -64,7 +63,7 @@ public override Task<SourcePackageDependencyInfo> ResolvePackage(PackageIdentity
                 try
                 {
                     // Retrieve all packages
-                    var repoPackage = V2Client.FindPackage(package.Id, legacyVersion, allowPrereleaseVersions: true, allowUnlisted: true);
+                    var repoPackage = GetSource().FindPackage(package.Id, legacyVersion, allowPrereleaseVersions: true, allowUnlisted: true);
```
yishaigalatzer, line 66: long line

src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
```diff
@@ -149,7 +148,45 @@ private SourcePackageDependencyInfo CreateDependencyInfo(IPackage packageVersion
...
+                    identity,
+                    deps,
+                    PackageExtensions.IsListed(packageVersion),
+                    _source);
+            }
+
+            return result;
+        }
+
+        private IPackageRepository GetSource()
```
yishaigalatzer, line 179: Should be called GetRespository() or GetClient()

src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
```diff
@@ -149,7 +148,45 @@ private SourcePackageDependencyInfo CreateDependencyInfo(IPackage packageVersion
...
+        {
+            var dataServiceRepo = V2Client as DataServicePackageRepository;
+
+            if (dataServiceRepo != null)
+            {
+                var sourceUri = new Uri(dataServiceRepo.Source);
+                dataServiceRepo = new DataServicePackageRepository(sourceUri);
+            }
+
+            return dataServiceRepo ?? V2Client;
```
yishaigalatzer, line 189: add a comment to the case where we return V2Client rather that dataServiceRepo

src/NuGet.Protocol.Core.v2/DependencyInfoResourceV2.cs
```diff
@@ -149,7 +148,45 @@ private SourcePackageDependencyInfo CreateDependencyInfo(IPackage packageVersion
...
+                    PackageExtensions.IsListed(packageVersion),
+                    _source);
+            }
+
+            return result;
+        }
+
+        private IPackageRepository GetSource()
+        {
+            var dataServiceRepo = V2Client as DataServicePackageRepository;
```
yishaigalatzer, line 181: instead of abbreviating just use `repository`

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -38,27 +39,29 @@ public DownloadResourceV2(V2Resource resource)
...
                 try
                 {
-                    var package = V2Client.FindPackage(identity.Id, version);
+                    var sourcePackage = identity as SourcePackageDependencyInfo;
+                    var dataServiceRepo = V2Client as DataServicePackageRepository;
 
-                    if (package != null)
+                    if (sourcePackage != null 
+                        && dataServiceRepo != null
+                        && sourcePackage.PackageHash != null
```
yishaigalatzer, line 51: nit: you can just use `sourcePackage?.PackageHash != null`

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -38,27 +39,29 @@ public DownloadResourceV2(V2Resource resource)
...
             return Task.Run(() =>
             {
-                var version = SemanticVersion.Parse(identity.Version.ToString());
+                token.ThrowIfCancellationRequested();
+
                 try
                 {
-                    var package = V2Client.FindPackage(identity.Id, version);
+                    var sourcePackage = identity as SourcePackageDependencyInfo;
+                    var dataServiceRepo = V2Client as DataServicePackageRepository;
```
yishaigalatzer, line 47: same as above: `repository`

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -38,27 +39,29 @@ public DownloadResourceV2(V2Resource resource)
...
-                            {
-                                return new DownloadResourceResult(package.GetStream(), new PackageFolderReader(directoryInfo));
-                            }
-                        }
-
-                        return new DownloadResourceResult(package.GetStream());
+                        // If this is a SourcePackageDependencyInfo object with everything populated 
+                        // and it is from an online source, use the machine cache and download it using the
+                        // given url.
+                        // If this info is not provided fallback to the old method.
```
yishaigalatzer, line 57: not clear from the comment where is the fallback. I assume it is the code after the else?

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -66,5 +69,211 @@ public DownloadResourceV2(V2Resource resource)
...
+        /// True if the given package matches hash
+        /// </summary>
+        private static bool MatchPackageHash(IPackage package, string hash)
+        {
+            var hashProvider = new CryptoHashProvider("SHA512");
+
+            return package != null && package.GetHash(hashProvider).Equals(hash, StringComparison.OrdinalIgnoreCase);
+        }
+
+        private static IPackage DownloadToMachineCache(
```
yishaigalatzer, line 198: I don't see cross process or thread safety in this code

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -66,5 +69,211 @@ public DownloadResourceV2(V2Resource resource)
...
+            IPackageRepository repository, 
+            CancellationToken token)
+        {
+            var version = SemanticVersion.Parse(identity.Version.ToString());
+            var dataServiceRepo = repository as DataServicePackageRepository;
+
+            if (dataServiceRepo != null)
+            {
+                // Clone the repo to allow for concurrent calls
+                var sourceUri = new Uri(dataServiceRepo.Source);
```
yishaigalatzer, line 143: This should probably be an extension method on the repo (or a method proper)

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -66,5 +69,211 @@ public DownloadResourceV2(V2Resource resource)
...
+                {
+                    return new DownloadResourceResult(downloadedPackage.GetStream());
+                }
+            }
+            else
+            {
+                var package = repository.FindPackage(identity.Id, version);
+
+                // Use a folder reader for unzipped repos
+                if (repository is UnzippedPackageRepository)
```
yishaigalatzer, line 170: Does this work with both
1. Flat nupkg folder
2. V3 nupkg structure (not unzipped)

src/NuGet.Protocol.Core.v2/project.json
```diff
@@ -27,7 +27,8 @@
     "net45": {
       "frameworkAssemblies": {
         "System.Collections.Concurrent": "",
-        "System.Runtime.Serialization": ""
+        "System.Runtime.Serialization": "",
+        "WindowsBase": ""
```
yishaigalatzer, line 31: How did WPF made it into the protocol layer?

src/NuGet.Protocol.Core.v3/Constants.cs
```diff
@@ -21,6 +21,7 @@ public static class ServiceTypes
     {
         public static readonly string VersionV2 = "/2.0.0";
         public static readonly string TypeVersion = "/3.0.0-beta";
+        public static readonly string Version300 = "/3.0.0";
```
yishaigalatzer, line 24: Why a new convention and not VersionV3 ?

src/NuGet.Protocol.Core.v3/DownloadResourceV3.cs
```diff
@@ -18,33 +18,81 @@ public class DownloadResourceV3 : DownloadResource
...
+        public DownloadResourceV3(HttpClient client, string flatContainerBaseUrl)
+            : this(client)
+        {
+            if (flatContainerBaseUrl == null)
             {
-                throw new ArgumentNullException("regResource");
+                throw new ArgumentNullException(nameof(flatContainerBaseUrl));
+            }
+
+            _flatContainerBaseUrl = flatContainerBaseUrl.TrimEnd('/');
```
yishaigalatzer, line 48: Where is the url coming from? Can it ever have space after the trailing `/` ?

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -101,53 +110,74 @@ public ServiceIndexResourceV3Provider()
...
                     {
-                        SemanticVersion version;
-                        if (SemanticVersion.TryParse((string)versionToken, out version) &&
-                            version.Major == 3)
-                        {
-                            index = new ServiceIndexResourceV3(json, utcNow);
-                        }
-                        else
+                        // wait up to a minute for the lock, if we cannot get it just retrieve the file again.
+                        release = await _semaphore.WaitAsync(TimeSpan.FromMinutes(1), token);
```
yishaigalatzer, line 122: Not sure I follow the logic of waiting a minute, what is the point?

src/NuGet.Protocol.Core.v2/project.json
emgarten, line 31: DataServicePackageRepository uses weak events.

src/NuGet.Commands/RestoreRequest.cs
emgarten, line 31: GetCoreV3 sets up the V3 resources, which for the project.json resolver includes v2 support since it uses a different provider that does not use nuget.core.

src/NuGet.Configuration/PackageSource/PackageSource.cs
emgarten, line 69: This follows the same pattern as IsHttp. IsLocal will only be hit for packages.config restores.

src/NuGet.Protocol.Core.Types/SourcePackageDependencyInfo.cs
emgarten, line 21: Many of the parameters are optional, is there a reason not to have these constructors?

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
emgarten, line 198: The v2 machine cache has never had mutexes like the global folder now does. It downloads to a tmp file and swaps the file after the download completes. It would be nice to have them if we wish to devote more cycles to NuGet.Core.

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
emgarten, line 143: I'll add a Clone method

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
emgarten, line 170: Yes, there are no changes for non-online v2 repos.

src/NuGet.Protocol.Core.v3/Constants.cs
emgarten, line 24: I'll clean these up

src/NuGet.Protocol.Core.v3/DownloadResourceV3.cs
emgarten, line 48: It comes from index.json. I normalize the ending since this code constructs urls, but I don't think it should go too much farther to massage the url into the correct form.

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
emgarten, line 122: It could be removed. The logic is that requesting index.json should be a fairly cheap operation, if for some reason another thread is taking over a minute to complete the call, just get it a second time.

src/NuGet.Commands/RestoreCommand.cs
emgarten, line 874: If not we should, this would break every where in the codebase.

src/NuGet.Configuration/PackageSource/PackageSource.cs
yishaigalatzer, line 69: :+1:

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
yishaigalatzer, line 122: In that case a minute is too long, 1-2 seconds should be sufficient

yishaigalatzer, line 122: In that case a minute is too long, 1-2 seconds would be more appropriate (so prevent unnecessary calls, but don't hold the user back).

emgarten, line 122: I'll update this.

src/NuGet.Protocol.Core.v2/DownloadResourceV2.cs
```diff
@@ -66,5 +69,211 @@ public DownloadResourceV2(V2Resource resource)
...
+            CancellationToken token)
+        {
+            var packageName = new PackageNameWrapper(package);
+            var version = SemanticVersion.Parse(package.Version.ToString());
+            IPackage newPackage = null;
+
+            FileInfo tmpFile = null;
+            FileStream tmpFileStream = null;
+
+            var downloadClient = new HttpClient(downloadUri)
```
deepakaravindr, line 212: I would add a comment that this is not the HttpClient from Microsoft.Net.Http package (or System.Net.Http.dll)

src/NuGet.Protocol.Core.v3/DownloadResourceV3.cs
```diff
@@ -18,33 +18,81 @@ public class DownloadResourceV3 : DownloadResource
...
+        /// <summary>
+        /// Download packages using the download url found in the registration resource.
+        /// </summary>
         public DownloadResourceV3(HttpClient client, RegistrationResourceV3 regResource)
+            : this(client)
         {
             if (client == null)
             {
-                throw new ArgumentNullException("client");
+                throw new ArgumentNullException(nameof(regResource));
```
deepakaravindr, line 31: Looks like you are checking if client is null and throw ArgNull for regResource. Perhaps, you need 2 of them

src/NuGet.Protocol.Core.v3/ServiceIndexResourceV3Provider.cs
```diff
@@ -55,34 +56,41 @@ public ServiceIndexResourceV3Provider()
...
                         }
-
-                        if (credentials == null)
+                        else
                         {
                             response.EnsureSuccessStatusCode();
                         }
                     }
-                    else
+                    catch (Exception)
```
deepakaravindr, line 91: Please remove this as it may be added by mistake

deepakaravindr:
Looks good to me

PR #134, opened 2015-09-28
PackagesConfigReader support for duplicate package id entries (part 1)

emgarten:
This change adds a flag on GetPackages to allow the packages.config reader to return entries with duplicate package ids for the solution level package restore scenario.

Package versions will always be verified, if two identical packages exist in packages.config an exception will be thrown.

By default duplicate ids are not allowed, this flag will only be set in the command line. When the vsix supports solution level packages again it will be used there also for the solution level config.

//cc @danliu @feiling @deepakaravindr @yishaigalatzer

### DIFF
--- a/src/NuGet.Packaging/PackagesConfigReader.cs
+++ b/src/NuGet.Packaging/PackagesConfigReader.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+using System.Linq;
 using System.Xml.Linq;
 using NuGet.Frameworks;
 using NuGet.Packaging.Core;
@@ -122,10 +123,20 @@ namespace NuGet.Packaging
         }
 
         /// <summary>
-        /// Reads all package node entries in the config
+        /// Reads all package node entries in the config.
+        /// If duplicate package ids exist an exception will be thrown.
         /// </summary>
-        /// <returns></returns>
         public IEnumerable<PackageReference> GetPackages()
+        {
+            return GetPackages(allowDuplicatePackageIds: false);
+        }
+
+        /// <summary>
+        /// Reads all package node entries in the config.
+        /// </summary>
+        /// <param name="allowDuplicatePackageIds">If True validation will be performed to ensure that 
+        /// only one entry exists for each unique package id.</param>
+        public IEnumerable<PackageReference> GetPackages(bool allowDuplicatePackageIds)
         {
             var packages = new List<PackageReference>();
 
@@ -191,18 +202,25 @@ namespace NuGet.Packaging
             }
 
             // check if there are duplicate entries
-            var packageIds = new HashSet<string>(StringComparer.OrdinalIgnoreCase);
-            var duplicates = new List<string>();
-            foreach (var package in packages)
+            IReadOnlyList<string> duplicates = null;
+
+            if (allowDuplicatePackageIds)
             {
-                if (packageIds.Contains(package.PackageIdentity.Id))
-                {
-                    duplicates.Add(package.PackageIdentity.Id);
-                }
-                else
-                {
-                    packageIds.Add(package.PackageIdentity.Id);
-                }
+                // Verify that no duplicate package identities
+                duplicates = packages.GroupBy(package => package.PackageIdentity, PackageIdentity.Comparer)
+                    .Where(group => group.Count() > 1)
+                    .Select(group => group.Key.ToString())
+                    .OrderBy(s => s, StringComparer.OrdinalIgnoreCase)
+                    .ToList();
+            }
+            else
+            {
+                // Verify no duplicate ids (this also ensures identities)
+                duplicates = packages.GroupBy(package => package.PackageIdentity.Id, StringComparer.OrdinalIgnoreCase)
+                    .Where(group => group.Count() > 1)
+                    .Select(group => group.Key.ToString())
+                    .OrderBy(s => s, StringComparer.OrdinalIgnoreCase)
+                    .ToList();
             }
 
             if (duplicates.Count > 0)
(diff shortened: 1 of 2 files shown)


test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
```diff
@@ -167,19 +167,108 @@ public void PackagesConfigReader_InvalidAllowedVersions()
...
+                  <package id=""test1"" version=""1.1""/>
+                </packages>");
+            var reader = new PackagesConfigReader(doc);
+
+            // Act
+            var ex = Assert.Throws<PackagesConfigReaderException>(() =>
+                reader.GetPackages(allowDuplicatePackageIds: true));
+
+            // Assert
+            Assert.Equal(ex.Message, "There are duplicate packages: test1.1.1.0");
```
danliu, line 271: is this message correct? why it says 1.1.1.0 now.

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
```diff
@@ -167,19 +167,108 @@ public void PackagesConfigReader_InvalidAllowedVersions()
...
+                  <package id=""test2"" version=""2.0""/>
+                </packages>");
+            var reader = new PackagesConfigReader(doc);
+
+            // Act
+            var ex = Assert.Throws<PackagesConfigReaderException>(() => 
+                reader.GetPackages(allowDuplicatePackageIds: true));
+
+            // Assert
+            Assert.Equal(ex.Message, "There are duplicate packages: test1.1.0.0");
```
danliu, line 230: error message is not correct/accurate

deepakaravindr:
:shipit:

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
emgarten, line 271: that's the package id and version

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
emgarten, line 230: It gives the full identity of the package since the error is about the id/version and not just the version. What is incorrect here?

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
yishaigalatzer, line 271: That's a bit misleading. Can we just name the package TestOne ?

src/NuGet.Packaging/PackagesConfigReader.cs
```diff
@@ -191,18 +202,25 @@ public IEnumerable<PackageReference> GetPackages()
...
-                if (packageIds.Contains(package.PackageIdentity.Id))
-                {
-                    duplicates.Add(package.PackageIdentity.Id);
-                }
-                else
-                {
-                    packageIds.Add(package.PackageIdentity.Id);
-                }
+                // Verify that no duplicate package identities
+                duplicates = packages.GroupBy(package => package.PackageIdentity, PackageIdentity.Comparer)
```
yishaigalatzer, line 210: The list of packages may be very long, and groupby could end up being too expensive. 

This can probably be simplified a whole lot, by just sorting by the package id and version outside the if. Then comparing each new Id to the next one.

Then just iterate the list to find duplicates.

yishaigalatzer:
:watch:

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
danliu, line 230: I understand the duplicate is test1, version 1.0.0 as the normalized version. But in the packages config itself it says 1.0.

Instead of using packageIdentity.ToString(), can you output something that matches what's shown on the packages.config file? So when user goes to packages.config to find the duplicate, it's easy for them to get what has been duplicated.

src/NuGet.Packaging/PackagesConfigReader.cs
danliu, line 210: +1. 
To test it, try with a packages.config with 200+ entries and see what the performance would be like.

test/NuGet.Packaging.Test/PackagesConfigReaderTests.cs
emgarten, line 230: Normalization makes it easier, for example in the test where there are two entries and they are non-normalized this makes it clear. We should normalize versions in all messages.

src/NuGet.Packaging/PackagesConfigReader.cs
```diff
@@ -122,11 +123,21 @@ public NuGetVersion GetMinClientVersion()
...
         {
+            return GetPackages(allowDuplicatePackageIds: false);
+        }
+
+        /// <summary>
+        /// Reads all package node entries in the config.
+        /// </summary>
+        /// <param name="allowDuplicatePackageIds">If True validation will be performed to ensure that 
+        /// only one entry exists for each unique package id.</param>
+        public IEnumerable<PackageReference> GetPackages(bool allowDuplicatePackageIds)
```
daazose, line 139: @yishaigalatzer are there plans to expose this as an option to the restore command?

yishaigalatzer, line 139: No it will always be true

Sent from my Windows Phone

src/NuGet.Packaging/PackagesConfigReader.cs
```diff
@@ -122,11 +123,21 @@ public NuGetVersion GetMinClientVersion()
...
 
         /// <summary>
-        /// Reads all package node entries in the config
+        /// Reads all package node entries in the config.
+        /// If duplicate package ids exist an exception will be thrown.
         /// </summary>
-        /// <returns></returns>
         public IEnumerable<PackageReference> GetPackages()
         {
+            return GetPackages(allowDuplicatePackageIds: false);
```
daazose, line 131: @yishaigalatzer, @emgarten - should this be set to true given the response below?

yishaigalatzer, line 131: No - that's the default behavior. There is a similar pr in nuget.package.management that covers the restore command

Sent from my Windows Phone

PR #137, opened 2015-09-30
Support random combinations of NuGet.config file names to support DNX

emgarten:
Adding support for alternative nuget.config casings. On non-windows machines directories will be checked for all three casings instead of just "NuGet.Config", on windows there should be no change.

Based on:
https://github.com/aspnet/dnx/commit/9dca281a9594eec292c0bb502e31535900a23223

https://github.com/NuGet/Home/issues/1427

//cc @pranavkm @yishaigalatzer @deepakaravindr @zhili1208 @MeniZalzman

### DIFF
--- a/src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
+++ b/src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
@@ -13,7 +13,6 @@ namespace NuGet.Configuration
     public class PackageSourceProvider : IPackageSourceProvider
     {
         private const int MaxSupportedProtocolVersion = 3;
-        private static Lazy<bool> _isMono = new Lazy<bool>(() => Type.GetType("Mono.Runtime") != null);
 
         private ISettings Settings { get; set; }
         private readonly IEnumerable<PackageSource> _providerDefaultPrimarySources;
@@ -84,7 +83,7 @@ namespace NuGet.Configuration
         {
 #if !DNXCORE50
             // Global default NuGet source doesn't make sense on Mono
-            if (_isMono.Value)
+            if (RuntimeEnvironmentHelper.IsMono)
             {
                 return Enumerable.Empty<PackageSource>();
             }
--- a/src/NuGet.Configuration/Utility/RuntimeEnvironmentHelper.cs
+++ b/src/NuGet.Configuration/Utility/RuntimeEnvironmentHelper.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.IO;
+
+namespace NuGet.Configuration
+{
+    internal static class RuntimeEnvironmentHelper
+    {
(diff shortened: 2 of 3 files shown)


src/NuGet.Configuration/Utility/RuntimeEnvironmentHelper.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.IO;
+
+namespace NuGet.Configuration
+{
+    internal static class RuntimeEnvironmentHelper
+    {
+        private static Lazy<bool> _isMono = new Lazy<bool>(() => Type.GetType("Mono.Runtime") != null);
+
+        // Checking the OS environment variable is most complete way to check this on dnxcore currently.
+        private static Lazy<bool> _isWindows = new Lazy<bool>(() =>
```
pranavkm, line 11: Think you could get away with doing eagerly evaluating this

`private static readonly bool _isWindows = Environment.GetEnvironmentVariable("OS")?.Equals("WINDOWS_NT") ?? false`

yishaigalatzer:
:shipit:

src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
```diff
@@ -84,7 +83,7 @@ private IEnumerable<PackageSource> LoadConfigurationDefaultSources(IEnumerable<P
         {
 #if !DNXCORE50
             // Global default NuGet source doesn't make sense on Mono
-            if (_isMono.Value)
+            if (RuntimeEnvironmentHelper.IsMono)
```
yishaigalatzer, line 86: we shouldn't check for Mono. We should check for a case sensitive file system.

yishaigalatzer, line 86: Not sure if there is a really good way to do it, but I assume that if you are on *nix you can rely on the pathseparator being forward slash. Alternatively the answer from http://stackoverflow.com/questions/430256/how-do-i-determine-whether-the-filesystem-is-case-sensitive-in-net is pretty heavy handed but can be done. Note that it is not accurate because it depends on the folder you are in.

emgarten, line 86: This was an existing check for mono, it's not related to the casing check. I just moved the helper method.

emgarten:
@yishaigalatzer this should go in 3.2.1

src/NuGet.Configuration/PackageSource/PackageSourceProvider.cs
yishaigalatzer, line 86: I see, why is this change part of this PR then?

yishaigalatzer:
@emgarten was this not merged yet? We should.

emgarten:
@yishaigalatzer it didn't have enough squirrels to make the jump. I will merge it now.

